PR #1, opened 2019-03-12
MEN-2397: First prototype for a dockerized deb builder

lluiscampos:
Hello!

I am still working on some improvements and optimizations, but I wanted to launch the PR already to get feedback (specially on the design level).

This repo will contain Docker image(s) and scirpt(s) to generate distribution packages for multiple architectures. For now, it only supports `.deb` packaging of the mender client for `armhf` (Raspberry Pi).

The idea is to create an image with all the dependencies, then install an script that will do the compile and package.

Thanks for your feedback!


Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM debian:stretch
+
+#TODO: Verify Go version for mender 2.0.0 beta release
+ARG MENDER_VERSION=2.0.0-beta
+ARG GOLANG_VERSION=1.11.5
+
+RUN apt-get update && apt-get install -y \
+    git wget \
+    build-essential debhelper devscripts \
```
lluiscampos: I am working on removing some dependencies (i.e. find the actual tools we need and not all these meta-packages)

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    liblzma-dev
+
+# TODO: test with gcc-linaro-6.3.1-2017.05-x86_64_arm-linux-gnueabihf.tar.xz !!
```
lluiscampos: I am working on this too

mender-deb-package
```diff
@@ -0,0 +1,107 @@
...
+%:
+	dh \$@
+
+override_dh_auto_build:
+	env CGO_CFLAGS="-I/usr/local/include" CGO_LDFLAGS="-L/usr/local/lib" \
+  GOOS=linux GOARCH=arm GOARM=5 CGO_ENABLED=1 \
+  make build
+
+override_dh_auto_install:
+	sed -i 's/prefix ?=/prefix="\$\${DESTDIR}"/' Makefile
```
lluiscampos: @kacf This is because inside mender client Makefile we use `prefix` only.

I couldn't find a lot of documentation, but it seems like debmake uses `DESTDIR` instead of `prefix` to pint to the destination directory. See [this link](https://www.debian.org/doc/manuals/debmake-doc/ch08.en.html#makefile-sh)

So we could leave this hack or modify the Makefile (so far I assume only the new integration tests are using it)

mender-deb-package
```diff
@@ -0,0 +1,107 @@
...
+EOF
+
+dpkg-buildpackage -aarmhf -us -uc -b
+
+if [ -d "/output" ]; then
+	for file in $(find ../ -maxdepth 1 -type f); do
+		cp ${file} /output
+	done
+else
+	echo "Warning: /output directory doesn't exist. Skipping file copy"
```
lluiscampos: Maybe we should convert this into an error? (A check at the start to not continue if there is no binded output directory)

lluiscampos:
@kacf @apodogrocki

lluiscampos:
@mirzak please have a look you too. Thanks!

Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+RUN git clone --progress https://github.com/raspberrypi/tools.git
+ENV CC "arm-linux-gnueabihf-gcc"
+ENV PATH "$PATH:/tools/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian-x64/bin"
```
apodogrocki: As already @mirzak noticed, why do you need raspberrypi tools at all? Is it only needed to get a toolchain to build Mender Client? If so, then you'd better take some more recent toolchain from e.g. linaro or even simply try to install it from packages: gcc-arm-linux-gnueabihf. 
Please take a look how both approaches are done in mender-convert: https://github.com/mendersoftware/mender-convert/commit/a231c92edde3c01d7902dde03772bad60e2fbc5b#diff-3254677a7917c6c01f55212f86c57fbfL38

kacf: Exactly my comment as well. You can try `apt install crossbuild-essential-armhf` perhaps.

README.md
```diff
@@ -0,0 +1,64 @@
+mender-dist-packages
+====================
+
+Mender is an open source over-the-air (OTA) software updater for embedded Linux devices. Mender comprises a client running at the embedded device, as well as a server that manages deployments across many devices.
+
+This repository contains mender-dist-packages, which is used to create distribution packages of the mender client software
+
+Currently only supports creation of a single `.deb` for `arm` arquitecture suitable for Raspberry Pi 3.
```
kacf: `s/arquitecture/architecture/`

README.md
```diff
@@ -0,0 +1,64 @@
...
+Create a directory output:
+
+```bash
+mkdir output
+```
+
+Create the deb package:
+
+```bash
+docker run --mount type=bind,source=$(pwd)/output,target=/output --privileged=true mender-dist-packages
```
kacf: * I think you don't need `--privileged` (is it because of the `--mount`?) if you use volume instead:
  ```
  docker run -v $(pwd)/output:/output mender-dist-packages
  ```
* I would add `--rm` as well, so that the container is disposed of afterwards.
* Would it make sense to put this in a `build-deb` script?

mender-deb-package
```diff
@@ -0,0 +1,107 @@
...
+cat << EOF > debian/rules
+#!/usr/bin/make -f
+
+%:
+	dh \$@
+
+override_dh_auto_build:
+	env CGO_CFLAGS="-I/usr/local/include" CGO_LDFLAGS="-L/usr/local/lib" \
+  GOOS=linux GOARCH=arm GOARM=5 CGO_ENABLED=1 \
+  make build
```
kacf: Weird indentation. Use tabs maybe?

mender-deb-package
kacf: Yes, agree.

mender-deb-package
kacf: I think we should definitely modify the Makefile. Just add

```
DESTDIR ?= /
prefix ?= $(DESTDIR)
```

This way the user can override either one.

lluiscampos: Good suggestion :+1:

README.md
lluiscampos: >     * I think you don't need `--privileged` (is it because of the `--mount`?) if you use volume instead:
>       ```
>       docker run -v $(pwd)/output:/output mender-dist-packages
>       ```
> 
Yes, it was because of the mount. Changed now
>     * I would add `--rm` as well, so that the container is disposed of afterwards.
>
Great, I didn't know about this flag!
>     * Would it make sense to put this in a `build-deb` script?
Soon we will need to modify this a bit to do several packages and probably to support different architectures, and then it would definitely make sense to wrap everything in an script. IMO for now (single manually created deb package) live with the instructions.

README.md
lluiscampos: Spanglish :+1:

Dockerfile
lluiscampos: Ok, I will leave it as it is.

mender-deb-package
lluiscampos: :+1:

Dockerfile
lluiscampos: I discussed it with @mirzak this morning.

Turns out that the problem is that my simulated Raspberry Pi is using ARMv6, so that when I tested with `gcc-arm-linux-gnueabihf` didn't work. But now I am using a real Raspberry Pi 3 (ARMv7) and can use `gcc-arm-linux-gnueabihf` :+1:

kacf: You should compile as ARMv5 though, because it is the most compatible. There are options for gcc to turn this on. This package may be installed on a variety of devices, so we should not assume ARMv7.

@mirzak: Feel free to correct me if you know more about the market share of each architecture, but this is the approach I used for our GRUB binary.

lluiscampos:
Pushed the new change set. Verified on a real Raspberry Pi 3 this time

mender-deb-package
```diff
@@ -0,0 +1,107 @@
...
+
+cat << EOF > debian/rules
+#!/usr/bin/make -f
+
+%:
+	dh \$@
+
+override_dh_auto_build:
+	env CGO_CFLAGS="-I/usr/local/include" CGO_LDFLAGS="-L/usr/local/lib" \
+	GOOS=linux GOARCH=arm GOARM=7 CGO_ENABLED=1 \
```
lluiscampos: @kacf I changed GOARM to 7. Is that correct? Should we just not specify it and go with the Golang default? On the Raspberry Pi 3 (ARMv7), it works for the three versions (5,6,7).

kacf: Yes, it works on RPi because ARMv7 is backwards compatible with ARMv5 and 6, but not vice versa. So an ARMv5 device will not be able to run ARMv7 binaries. Hence why I think we should compile with that.

Dockerfile
lluiscampos: Discussed with @mirzak on chat.

Turns out that `gcc-arm-linux-gnueabihf` does only support ARMv7. Also Debian upstream seems to support only v7 and newer.

I tried with GCC flag `-march=armv5te` but the toolchain seemed to ignore it. When manually compiling a hello world program with the flag, I get an error saying that it is  unimplemented.

So my suggestion is to leave gcc compilation for ARMv7.

kacf: Ok, then we have no choice, I guess.

mender-deb-package
lluiscampos: And I also made experiments setting GOARM=5, =6, =7... When I read back the binaries, I always get the same result, that it is compiled for v7. See below the output of readelf from a build with GOARM=5

# readelf -A mender_arm5

mender_arm5
Attribute Section: aeabi
File Attributes
  Tag_CPU_name: "7-A"
  Tag_CPU_arch: v7
  Tag_CPU_arch_profile: Application
  Tag_ARM_ISA_use: Yes
  Tag_THUMB_ISA_use: Thumb-2
  Tag_FP_arch: VFPv3-D16
  Tag_ABI_PCS_wchar_t: 4
  Tag_ABI_FP_rounding: Needed
  Tag_ABI_FP_denormal: Needed
  Tag_ABI_FP_exceptions: Needed
  Tag_ABI_FP_number_model: IEEE 754
  Tag_ABI_align_needed: 8-byte
  Tag_ABI_enum_size: int
  Tag_ABI_VFP_args: VFP registers
  Tag_CPU_unaligned_access: v6

lluiscampos: I leave it with GOARM=5, though

lluiscampos:
See new commits as well.

mender-deb-package
kacf: Yes, but I think this is because we use `CGO_ENABLED=1`. It is GCC that does the linking, not Golang, so it boils down to whether GCC supports it. The actual code that Golang generates may still be ARMv5, but that doesn't help if the binary is wrapped in an ARMv7 header.

lluiscampos: Makes sense! Thank you for the clarification.

Dockerfile
```diff
@@ -4,11 +4,12 @@ FROM debian:stretch
 ARG MENDER_VERSION=2.0.0-beta
 ARG GOLANG_VERSION=1.11.5
 
-RUN apt-get update && apt-get install -y \
+RUN dpkg --add-architecture armhf && \
```
kacf: :+1:

PR #3, opened 2019-03-27
Prepare deb package for ARMv6 devices (Raspberry Pi Zero W)

lluiscampos:
Ported the toolchain change from mender-convert to the deb package, so
that we are consistent in the ARM support accross the mender offering.

Along the way:
* Build liblzma from source instead of using the Debian upstream pkg
* Hardcode deb package dependencies instead of using Debian helper
* Compile go code for ARM6 (to be consistent across mender products)
* Clean debug prints

Changelog: Title

### DIFF
--- a/mender-deb-package
+++ b/mender-deb-package
@@ -38,13 +38,6 @@ if [ ! -d "/output" ]; then
   show_help_and_exit
 fi
 
-#Print misc env variables for debug
-echo "WORKDIR="`pwd`
-echo "PATH="$PATH
-echo "CC="$CC
-echo "GOPATH="$GOPATH
-echo "version=$version"
-
 mkdir debian && mkdir debian/source
 
 #TODO: find a way to take this from the Changelog itself
@@ -64,7 +57,7 @@ Build-Depends: debhelper (>= 9)
 
 Package: mender
 Architecture: armhf
-Depends: \${shlibs:Depends}, \${misc:Depends}
+Depends: libc6 (>= 2.12), liblzma5 (>= 5.1)
 Description: mender client
  Mender is an open source over-the-air (OTA) software updater for embedded Linux devices.
 EOF
@@ -78,7 +71,8 @@ cat << EOF > debian/rules
 	dh \$@
 
 override_dh_auto_build:
-	env GOOS=linux GOARCH=arm GOARM=5 CGO_ENABLED=1 make build
+	make CGO_CFLAGS="-I${LIBLZMA_INSTALL_PATH}/include" CGO_LDFLAGS="-L${LIBLZMA_INSTALL_PATH}/lib" \
+        CC=$CC GOOS=linux GOARCH=arm GOARM=6 CGO_ENABLED=1 build
 
 override_dh_auto_test:
 	true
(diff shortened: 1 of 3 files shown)


mender-deb-package
```diff
@@ -64,7 +57,7 @@ Build-Depends: debhelper (>= 9)
 
 Package: mender
 Architecture: armhf
-Depends: \${shlibs:Depends}, \${misc:Depends}
+Depends: libc6 (>= 2.12), liblzma5 (>= 5.1.1)
```
lluiscampos: I have set these dependencies based on what Raspbian upstream has today as latest.

However, we are building with libc6 2.27, and 5.2.4. Will that be a problem?

It works on my Raspberry Pi 3...

lluiscampos:
@apodogrocki @kacf can you please have a look at this?

mender-deb-package
```diff
@@ -78,7 +71,8 @@ cat << EOF > debian/rules
 	dh \$@
 
 override_dh_auto_build:
-	env GOOS=linux GOARCH=arm GOARM=5 CGO_ENABLED=1 make build
+	env CGO_CFLAGS="-I${LIBLZMA_INSTALL_PATH}/include" CGO_LDFLAGS="-L${LIBLZMA_INSTALL_PATH}/lib" \
+        GOOS=linux GOARCH=arm GOARM=6 CGO_ENABLED=1 make build
```
apodogrocki: Don't you need: CC=$CC \

lluiscampos: CC It is already in the environment, so here is just passed to make. Not strictly needed (neither in mender-convert recipe)

mender-deb-package
kacf: I couldn't find any specific information about versioning of either component, but I think libc is extremely stable, and it looks like liblzma is using semantic versioning, so we can at least drop the last component (patch versions should be both forwards and backwards compatible).

But this is one case where I would consider not using the latest debian to compile, but rather an older one. That might land you on older versions of both of them.

mender-deb-package
kacf: I think in general it's recommended to drop `env` and put `make` first instead, because variables in the environment don't override variables in the makefile, but variables given directly to make do. In this case @apodogrocki would be right: You do need `CC=$CC`.

lluiscampos: Thanks. I understand now :)

I will rework

mender-deb-package
lluiscampos: Mmmm. Good idea, so using an older build machine will ensure that the latest Raspbian (upstream) has newer versions of the dependencies. I will fix it to build from an older build host

Did I understand correctly?

And yes, I will drop the patch part on the versioning.

mender-deb-package
lluiscampos: Done :+1:

mender-deb-package
lluiscampos: As discussed in person; this conversation is really a misunderstanding. The "build" versions are not from Debian upstream, but from the downloaded bootlin toolchain (libc6) or the dependencies source code (liblzma).

Ironically, this PR started with the intention of aligning mender-convert and the .deb package support wrt ARM. Hence, we should not change here the toolchain nor liblzma source we are using to compile.

Bottom line: we know both dependencies are very stable and we know "it works". So we will leave it as it is. Objections?

lluiscampos:
I added also the docker build + deb build in Travis.

@apodogrocki @kacf approve? So that I can generate and upload the final .deb package

mender-deb-package
kacf: :+1:

PR #7, opened 2019-05-10
MEN-2396: Package client integ tests

lluiscampos:
This PR brings integration tests to the repo. For now it tests the only existing package, `mender-client`.

I tried to keep the tests simple, and left for whenever required future refactoring. Also the design ended up being a big messy, as helpers.py and setup.py are not very well differentiated.

@mzedel Please have a look on the "dockerizing" part.
@kacf Please have a look at the rest.

All your feedback is welcomed!


tests/test_package_client.py
```diff
@@ -36,7 +47,30 @@ def test_install_package(self, setup_tester_ssh_connection):
...
+            setup_tester_ssh_connection.run('test -x {invent}'.format(invent=inventory_path), hide=True)
+        for identity in self.expected_indentity_files:
+            identity_path = os.path.join("/usr/share/mender/identity", identity)
+            setup_tester_ssh_connection.run('test -x {ident}'.format(ident=identity_path), hide=True)
+        setup_tester_ssh_connection.run('test -d /etc/mender', hide=True)
+        setup_tester_ssh_connection.run('test -f /etc/mender/mender.conf', hide=True)
+        setup_tester_ssh_connection.run('test -f /etc/mender/mender.conf.demo', hide=True)
+        setup_tester_ssh_connection.run('test -f /etc/mender/mender.conf.production', hide=True)
+        setup_tester_ssh_connection.run('cmp /etc/mender/mender.conf.production /etc/mender/mender.conf', hide=True)
+        setup_tester_ssh_connection.run('test -f /lib/systemd/system/mender.service', hide=True)
```
kacf: Tip: I think you can do:

```
with settings(hide=True):
    setup_tester_ssh_connection.run('test -x /usr/bin/mender')
```

to avoid having to use `hide=True` everywhere.

tests/setup.py
```diff
@@ -67,7 +80,7 @@ def _probe_ssh_connection(conn):
                 ready = True
 
         except SSHException as e:
-            if not str(e).endswith("Connection reset by peer"):
+            if not (str(e).endswith("Connection reset by peer") or str(e).endswith("Error reading SSH protocol banner")):
```
kacf: Is the timeout 20 seconds? I would consider adding quite a bit more here. VMs with horrible performance is a quite common occurrence, so should have some leeway.

tests/docker-files/prepare-raspbian-img.sh
```diff
@@ -0,0 +1,50 @@
...
+set -e
+
+sudo true
+
+currdir=$(pwd)
+scriptdir=$(cd `dirname $0` && pwd)
+workdir=$(mktemp -d)
+
+#TODO: Verify image with md5sum
+if [ -f 2019-04-08-raspbian-mender-testing.img ]; then
```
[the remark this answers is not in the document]
lluiscampos: Yes, of course.

tests/pytest.ini
```diff
@@ -0,0 +1,4 @@
+# Disable stout capture, as it conflicts with Fabric
+#  see https://github.com/pytest-dev/pytest/issues/1585
+[pytest]
+addopts = --capture=no
```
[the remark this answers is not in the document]
lluiscampos: I didn't know if it is a problem specific to Fabric 2 or not. I can add it to other repos next time I submit a PR there.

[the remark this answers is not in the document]
lluiscampos: One question, though.

As you have seen, I added pytest.ini both to the root directory and to tests, so that you could potentially run pytest from both of them. The error message you when when running pytest without `-s` is quite hard to understand, so I preferred to make sure we don't run into the problem.

What do you think about this?

tests/test_package_client.py
```diff
@@ -0,0 +1,157 @@
...
+@pytest.fixture(scope="class")
+def standard_setup():
+    output = subprocess.check_output("docker run --rm --network host -d mender-dist-packages-testing", shell=True)
+    docker_container_id = output.decode("utf-8").split("\n")[0]
+
+    ready = wait_for_raspbian_boot(docker_container_id)
+
+    assert ready, "Raspbian QEMU image did not boot. Aborting"
+    yield
+    subprocess.check_output("docker stop {}".format(docker_container_id), shell=True)
```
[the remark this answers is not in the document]
lluiscampos: It is well explained here: https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code

We do this also in integration repo, using the deprecated `yield_fixture` decorator. This is something I think we will clean up when migrating to Python 3 / Fabric 2, as we can use the time to update our pytest version there.

tests/test_package_client.py
lluiscampos: Great! Thanks

tests/pytest.ini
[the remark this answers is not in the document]
kacf: Hmm, I think that pytest will collect (and run) Python code in all sub directories of the folder where you start it, so it may have unintended side effects if we advertise that it can be run outside the test folder. I'm not strictly opposed, but I think my vote would be to keep this in the test directory and always run it from there.

tests/setup.py
lluiscampos: I can add more, sure.

But just for you to get the full picture: there is already a 3 minutes timeout for the system to boot; this 20 seconds extra is only from the image reaching the login prompt to the ssh server being ready.

tests/test_package_client.py
lluiscampos: So the `with settings...` is a deprecated feature from Fabric 1 to Fabric 2, see http://www.fabfile.org/upgrading.html#configuration

I added the hide=True to the connection config, so that we nevertheless avoid repeating it in every `run` call.

tests/pytest.ini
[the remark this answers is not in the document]
lluiscampos: I agree.

I will let the test framework fail if you run it from the root repo directory.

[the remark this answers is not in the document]
lluiscampos: Oh! I just realized that I actually didn't commit the extra pytest.ini, so it was only on my local copy. Nothing to do there, then.

lluiscampos:
@kacf New change set

tests/docker-files/prepare-raspbian-img.sh
```diff
@@ -5,41 +5,78 @@
 
 set -e
 
+show_help_and_exit() {
+  cat << EOF
+Usage: $0 raspbian-version
+
+Aarguments:
```
kacf: "Aah", typo! :-)

lluiscampos:
@kacf Updated.

Modified also `Dockerfile` following comments from @mzedel

kacf:
Btw, I might shamelessly steal the test framework for use in https://github.com/mendersoftware/mender-demo-artifact as well! ;-)

tests/docker-build
```diff
@@ -0,0 +1,15 @@
...
+
+set -e
+
+raspbian_version="2019-04-08"
+
+scriptdir=$(cd `dirname $0` && pwd)
+
+(cd ${scriptdir}/docker-files && ./prepare-raspbian-img.sh ${raspbian_version})
+
+docker build --no-cache --build-arg raspbian_version=${raspbian_version} -t mender-dist-packages-tester ${scriptdir}/docker-files/
```
kacf: Btw, why `--no-cache`?

lluiscampos: It sholdn't be! I was experimenting locally and commit that by mistake in one of the rebases. Good catch, I'll fix

lluiscampos:
> Btw, I might shamelessly steal the test framework for use in https://github.com/mendersoftware/mender-demo-artifact as well! ;-)

This could be interesting to discuss. Maybe there is room for an independent repo with the framework?

kacf:
> This could be interesting to discuss. Maybe there is room for an independent repo with the framework?

Yes, that struck me as well. But I think I will experiment a bit first. It is possible that I will substantially modify some parts, and the framework is not that big. TBD.

lluiscampos:
Sounds good, let's experiment and learn about our needs first :)

PR #11, opened 2019-09-17
Upgrade to Mender client 2.1.0.

kacf:
Changelog: Title


kacf:
Thanks, fixed.

PR #12, opened 2019-09-18
MEN: 2728: ARM cross compile: align dependencies across components

oleorhagen:
Changelog: Move from locally compiled lzma library, to relying on the
Debian package.

### DIFF
--- a/mender-deb-package
+++ b/mender-deb-package
@@ -72,8 +72,7 @@ cat << EOF > debian/rules
 	dh \$@
 
 override_dh_auto_build:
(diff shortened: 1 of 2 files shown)


mender-deb-package
```diff
@@ -72,7 +72,7 @@ cat << EOF > debian/rules
 	dh \$@
 
 override_dh_auto_build:
-	make CGO_CFLAGS="-I${LIBLZMA_INSTALL_PATH}/include" CGO_LDFLAGS="-L${LIBLZMA_INSTALL_PATH}/lib" \
+	make CGO_CFLAGS="-idirafter/usr/include" CGO_LDFLAGS="-L/usr/lib/arm-linux-gnueabihf/" \
```
oleorhagen: I went for the compiler flag solution. However, for now `arm-linux-gnueabihf` is hardcoded. Does it need to be variabelized?

PR #13, opened 2019-09-25
Check in static debian files

lluiscampos:
```
commit e3ccdd03999d76e10a4336fa2afe4c678336eda9
Date:   Wed Sep 25 14:06:52 2019 +0200

    Check in static debian/ files instead of autogenerate them
    
    Except for the changelog, which we keep autogenerating to not modify the
    release procedure for now.
    
    This will allow for easier developement of the deb pacakging and ease
    the future merge with the official Debian recipes.
    
    Changelog: None
    
commit 09728f5bd43e9a5a724eeb5125272a86f38eacdc
Date:   Wed Sep 25 14:07:39 2019 +0200

    Update the build image to use latest Debian stable (buster)
    
    Changelog: None
    
```


Dockerfile
```diff
@@ -1,5 +1,4 @@
-FROM debian:stretch
-
+FROM debian:buster
```
kacf: Is this a good idea? I would argue that the package should be built from the oldest supported debian, not the newest. New releases are typically backwards compatible, but old releases are not forwards compatible.

lluiscampos: In this case we are using a custom toolchain and installing golang from upstream, so it is only liblzma-lib that we are using from Debian.

But anyway, in general terms I agree and we should get it as a general rule: to follow one stable Debian release behind the latest one.

I am dropping this commit.

lluiscampos:
@kacf Does the rest look ok?

PR #15, opened 2019-10-24
MEN-2426, MEN-2790 and QA-113: mender setup and related tasks

lluiscampos:
The merge of this PR is blocked by https://github.com/mendersoftware/mender/pull/436, but the changes can be reviewed.

Best reviewed commit by commit :)

This PR solves three tickets:
- https://tracker.mender.io/browse/MEN-2426
- https://tracker.mender.io/browse/MEN-2790
- https://tracker.mender.io/browse/QA-113


debian/control
```diff
@@ -7,7 +7,7 @@ Build-Depends: debhelper (>= 9)
 
 Package: mender-client
 Conflicts: mender
-Architecture: armhf
+Architecture: any
```
kacf: Is this overridden by the `dpkg-buildpackage` flags, or did you include this by mistake?

.gitlab-ci.yml
```diff
@@ -71,18 +70,28 @@ test:acceptance:
...
+  stage: publish
+  image: debian:buster
+  dependencies:
+    - build
+  before_script:
+    - apt update && apt install -yyq awscli
+  script:
+    - echo "Publishing ${MENDER_VERSION} packages to S3"
+    - deb_version=$(cat output/mender-client-deb-version)
+    # For master packages, the Debian version is "0.0~[iso-date].[git-hash]-1" and we make a copy named "master" for GUI to use
```
kacf: Under which circumstances is the GUI using this and not the official package?

debian/postinst
```diff
@@ -0,0 +1,28 @@
...
+            --demo=false \
+            --hosted-mender \
+            --tenant-token "Paste your Hosted Mender token here" \
+            --update-poll 1800 \
+            --inventory-poll 28800 \
+            --retry-poll 300
+
+    # Run the interactive setup wizard
+    else
+        mender setup
```
kacf: I thought there was a standard Debian frontend we should use, and then populate the values above? Did I misunderstand?

debian/postinst
```diff
@@ -26,3 +26,5 @@ if [ ! -f /etc/mender/mender.conf ]; then
     # Initialize artifact_info; should be removed after MEN-2585
     echo "artifact_name=unknown" > /etc/mender/artifact_info
 fi
+
+#DEBHELPER#
```
kacf: What's this? Special debian magic?

Dockerfile
```diff
@@ -35,4 +35,5 @@ RUN git checkout $MENDER_VERSION
 ENV mender_version $MENDER_VERSION
 COPY mender-deb-package /usr/local/bin/
 COPY debian/* debian/
+RUN cp support/mender.service debian/mender-client.service
```
kacf: Are `.service` files in `debian` auto-handled?

Also I'm wondering, does this automatically handle stopping the service when upgrading or removing? That's one thing that has been bugging me, that when you remove the package, the daemon still runs afterwards. Maybe you could have a look at it while you're at it.

debian/postrm
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+
+if [ "$1" == "purge" ]; then
+    rm /etc/mender/mender.conf
+    rm -f /etc/mender/artifact_info
+fi
```
kacf: Yes!

debian/postinst
lluiscampos: Yes: special Debian magic.

It took me a bit to figure it out, but this will make the debhelper insert here all other autogenerated postinst/postrm. For our case, this meant the all the handling of the systemd service when installing, removing or upgrading.

Dockerfile
lluiscampos: Related to the above.

Yes, it is handled automatically. For an update, for example, it will just stop, upgrade, start. For an install it will enable and start. For a remove stop + remove.

This is the case already for upstream Debian package but we were avoiding this by manually overriding the systemd targets in `rules`. We had it like this previously because we didn't have a way to configure the client.

debian/control
lluiscampos: It is the right thing to have for a multi arch package (and it is the same we have in Debian upstream recipe).

Basically it tells apt-get (if we put this on a repo) that the package will be available for any architecture, so then it will pick the one that matches the host architecture

Note that this is different that writing `all`, which means that the package is architecture agnostic (for example and script).

lluiscampos: If you are curios, ref https://www.debian.org/doc/debian-policy/ch-controlfields.html#architecture

.gitlab-ci.yml
lluiscampos: Maybe I am misunderstanding your question. What do you mean by official package? The official Debian packages? That ones are never used in the GUI, we only serve our own ones.

Then, a master build will make the GUI serve a link of the form `.../master/dist-packages/debian/mender-client_master-1_armhf.deb`, and that is why we make the copy. So in a way this "master" really points to the latest of all the `0.0~git...`

debian/rules
```diff
@@ -5,15 +5,13 @@
...
 override_dh_auto_build:
 	make CGO_CFLAGS="${CGO_CFLAGS}" CGO_LDFLAGS="${CGO_LDFLAGS}" CC=${CC} \
-	GOOS=linux GOARCH=arm GOARM=6 CGO_ENABLED=1 build
+	GOOS=linux CGO_ENABLED=1 build
 
-override_dh_auto_test:
-	true
+override_dh_auto_install:
+	# All install targets except install-bin and install-systemd
+	DESTDIR=debian/mender-client make install-bin install-identity-scripts install-inventory-scripts install-modules
```
lluiscampos: @kacf This is something I wanted to discuss. After mender-convert v2, I think we will be using the "make install" target only here for the deb package. So we might very well just modify it there removing the install-conf and install-systemd

lluiscampos: (This also made me realize that the comment is wrong, `install-bin` is there, what is missing is `install-conf`)

debian/postinst
lluiscampos: This might be something not very well communicated by my side. I brought it up in an stand-up and the JIRA ticket is updated.

ref https://tracker.mender.io/browse/MEN-2426

Basically I decided that the `mender setup` wizard is too involved and I didn't want to copy paste all the prompt texts and flow logic into `debconf` requirements. The Debian way is to have templates for all the questions, and then you write a script to handle all the logic (i.e. moving the user from one template to the next). So it is basically copying Alf work and I foresee maintenance burden here.

I have a prototype for this in a branch, so it is still open for discussion.

It might be accepted in Debian, as we are still providing a non interactive mode with defaults. I'll see if I can get some hints on this from Debian community.

.gitlab-ci.yml
kacf: By official I mean our own package, but the versioned one. I mean, where do we want to serve the master package and not the latest stable one? The former is not tested by us.

debian/rules
kacf: I agree with `install-conf`, since we now have a replacement in `mender setup`, but I think `install-systemd` should stay. I think the `make install` stage should be modeled after what a user installing from source would use, not what our scripts use.

.gitlab-ci.yml
lluiscampos: @mzedel Please see above for context, as I am not confident on my understanding. Will the GUI link to the mender-client master package under some circumstances?

mzedel: The .deb package `master` will be linked to in the help section and onboarding, unless a different mender version is defined - FWIW it used to just always be master.

debian/rules
lluiscampos: Done in https://github.com/mendersoftware/mender/pull/443

lluiscampos:
Changed the base branch of this PR to merge to a feature branch instead of master (tests still pending).

PR #16, opened 2019-10-28
MEN-2419: Integration test for interactive install

lluiscampos:
```
commit 2244db2b5e68476cd31be622b7a9ae8f11e56788
Date:   Mon Oct 28 15:35:27 2019 +0100

    Clean-up device_type file on purge in postrm step.
    
    This file is created by mender setup in postinst step, so it should be
    cleaned up accordignly on purge in postrm step.
    
    Changelog: None
    
commit e531df5f7e9e7b1785cf3ffd9033f5272171e260
Date:   Mon Oct 28 15:40:32 2019 +0100

    Rework current tests to use the new package in non-intreactive mode
    
    This commit removes the extra configuration steps and the enable/start
    of systemd service, as these are now handled on the deb pacakge install
    and remove logic.
    
    The installation of the pacakge is done in non-interactive mode, meaning
    using the default configuration for the Mender client.
    
    Changelog: None
    
commit bf1a647cfe61aa9afbf6a2437c99ef453ef8672c
Date:   Mon Oct 28 16:46:23 2019 +0100

    Add new test for the interactive installation of the client package
    
    Refactor the current code into a base class so that boths cases
    (interactive and non interactive) can share all the device checks.
    
    Changelog: None
    
commit bed1e13bd159d237f36e526123796aea98940adb
Date:   Tue Oct 29 16:41:12 2019 +0100

    Fixes in deb recipe to adapt to latest changes in mender master
    
    Basically:
    - Rename of parameter --hosted-mender to --mender professional
    - Partial restore of install-conf in Makefile (with artifact_info file)
    - Addition of demo certificate with install-examples Makefile target
    
    Changelog: None
    
```

### DIFF
--- a/debian/postrm
+++ b/debian/postrm
@@ -4,7 +4,6 @@ set -e
 
 if [ "$1" == "purge" ]; then
     rm /etc/mender/mender.conf
-    rm -f /etc/mender/artifact_info
     rm -f /var/lib/mender/device_type
 fi
 
--- a/debian/postrm
+++ b/debian/postrm
@@ -5,6 +5,7 @@ set -e
 if [ "$1" == "purge" ]; then
     rm /etc/mender/mender.conf
     rm -f /etc/mender/artifact_info
+    rm -f /var/lib/mender/device_type
 fi
 
 #DEBHELPER#
--- a/tests/test_package_client.py
+++ b/tests/test_package_client.py
@@ -21,10 +21,7 @@ import re
 from helpers import package_filename, upload_deb_package
 from mender_test_containers.helpers import *
 
-class TestPackageMenderClientDefaults():
-    """Tests instalation, setup, start, removal and purge of mender-client deb package with
-    in non-interactive method (i.e. default configuration).
-    """
+class PackageMenderClientChecker():
 
     expected_update_modules = ["deb", "directory", "rpm", "script", "single-file"]
     expected_inventory_files = ["mender-inventory-bootloader-integration",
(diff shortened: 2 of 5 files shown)


debian/postrm
```diff
@@ -5,6 +5,7 @@ set -e
 if [ "$1" == "purge" ]; then
     rm /etc/mender/mender.conf
     rm -f /etc/mender/artifact_info
+    rm -f /var/lib/mender/device_type
```
kacf: :+1:

lluiscampos:
> I haven't looked at the test yet. Is this WIP or ready to review?

It is ready for review.

I see that the test error is timing related. I will modify the logic for a more robust check of `journalctl` logs, but the overall test will not change.

lluiscampos:
> I see that the test error is timing related. I will modify the logic for a more robust check of `journalctl` logs, but the overall test will not change.

It should be better now.

tests/test_package_client.py
```diff
@@ -113,56 +98,42 @@ def test_start_client(self, setup_tester_ssh_connection):
...
+        setup_tester_ssh_connection.run('test ! -f /lib/systemd/system/mender-client.service')
+        setup_tester_ssh_connection.run('test ! -f /etc/systemd/system/multi-user.target.wants/mender-client.service')
         setup_tester_ssh_connection.run('test -d /etc/mender')
+        setup_tester_ssh_connection.run('test -f /etc/mender/mender.conf')
         setup_tester_ssh_connection.run('test -d /var/lib/mender/')
+        setup_tester_ssh_connection.run('test -f /var/lib/mender/device_type')
+
+        result = setup_tester_ssh_connection.run('sudo journalctl -u mender-client --no-pager')
+        assert "Stopping Mender OTA update service..." in result.stdout
+        assert "Stopped Mender OTA update service." in result.stdout
```
kacf: Would it make sense with a negative `pgrep mender` check here as well?

tests/test_package_client.py
```diff
@@ -97,35 +91,108 @@ def test_install_configure_start(self, setup_tester_ssh_connection, mender_dist_
...
+        # Are you connecting this device to Hosted Mender? [Y/n] n
+        # Do you want to run the client in demo mode? [Y/n] y
+        # Set the IP of the Mender Server: [127.0.0.1] 1.2.3.4
+        result = setup_tester_ssh_connection.run('sudo dpkg -i ' + package_filename(mender_dist_packages_versions["mender-client"]) + """ <<STDIN
+rasbperrytest
+y
+n
+y
+1.2.3.4
+STDIN""")
```
kacf: Should the resulting values in `/etc/mender/mender.conf` be checked?

tests/test_package_client.py
lluiscampos: Good idea, I will add it.

tests/test_package_client.py
lluiscampos: There are unit tests for this already in `mender` code base that check the actual contents.

Not having an easy way to launch builds on these repos with two PRs, I think that testing here too much will eventually become problematic.

I will add a basic check for IP addr and server url, but not the full file.

PR #19, opened 2019-12-04
Split Debian recipes to be able to build old Mender clients

lluiscampos:
Split Debian recipes to be able to build old Mender clients

This will allow us to still build 2.1.x, which at the moment is broken.

The design is to keep debian-X.X.x directories with the recipes for the
different branches of the mender client; debian-master representing the
latest recipe compatible with latest Mender Makefile.

Current recipe is moved to -master, and recipe from commit efdd73a is
taken as -2.1.x reference.

Dockerfile will copy all debian-* recipes, and then the build script
will decide which one to use for the actual build.


lluiscampos:
@mzedel Could you have a look at this one ASAP? I need to merge it before next release candidate build. The design was agreed with @kacf.

PR #20, opened 2019-12-04
Fix tag detection regex to accept build tags

lluiscampos:
Fix tag detection regex to accept build tags, tags can either be like 1.2.3 or 1.2.3-build4.


lluiscampos:
> LGTM, but those `[1-9]` parts really should include 0.

Updated.

Thanks @olehermanse for a quick but effective review!

PR #21, opened 2019-12-04
Allow integration tests failures on Pipeline

lluiscampos:
Allow integration tests failures on Pipeline

This is the only way (at the moment) to publish the deb packages for old
Mender client (i.e. 2.1.x).

A proper fix will be discussed and put in place with QA-123.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -28,7 +28,10 @@ build:
     paths:
       - output/*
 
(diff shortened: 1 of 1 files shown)


lluiscampos:
This is a temporary workaround that we need for current releases. A JIRA ticket has been created to discuss the proper fix.

Merging.

PR #23, opened 2019-12-05
Rename Hosted Mender setup flag and set default device type

alfrunes:
Made device type default to hostname in postinstall script.

changelog: none

Signed-off-by: Alf-Rune Siqveland <alf.rune@northern.tech>


lluiscampos:
This PR depends on https://github.com/mendersoftware/mender/pull/457

And the tests need to be updated also

alfrunes:
@lluiscampos I don't think the tests makes explicit use of the `--mender-professional` flag, I think the tests are failing because the quiet flag added in mendersoftware/mender#457 isn't merged yet.

PR #24, opened 2019-12-23
QA-134: Enable python3 formatting Gitlab pipeline

oleorhagen:
Changelog: None


kacf:
Shouldn't this and #25 be together?

oleorhagen:
> Shouldn't this and #25 be together?

I figured I'll split them up. One for adding the tests, and this one for formatting the files.

kacf:
But why is this passing then?

oleorhagen:
Because it only checks files which are added. Not existing files. 

This was a conscious design decision made when creating the pipeline. You disagree with the approach?

oleorhagen:
@kacf

kacf:
> This was a conscious design decision made when creating the pipeline. You disagree with the approach?

Yeah, kind of. Why not check all the code? I doesn't seem to add any benefit, and makes it difficult to run it locally in a predictable manner. I would just make a complete check for all the files.

oleorhagen:
Hmm, idea was that a PR should not fail on code that is unrelated to what you have committed yourself. If we are to change the approach, this has to be done on mendertesting though. 

See: 
https://github.com/mendersoftware/mendertesting/blob/master/check_python_code_format

kacf:
> Hmm, idea was that a PR should not fail on code that is unrelated to what you have committed yourself.

It never will, if all code is always checked.

> See:
> https://github.com/mendersoftware/mendertesting/blob/master/check_python_code_format
> 
> And code has to be filtered somehow. So it was either this approach, or adding a .pyproject.toml file with the necessary ignores to each project.
> 
> Another reason to keep it filtered like this, is that it seperates upgrading the format tool from other PR's. So if the tool is upgraded, all unrelated PR's will start failing.

This is why mendertesting has always been vendored before. Only when upgrading it in each repository can it potentially break. I can certainly see the benefits of running it non-vendored though, and with the number of repositories we have it makes sense. I think we just need to take a lot of care if we upgrade it.

> I don't understand the 'running it locally' critique? Why would it make it harder?

This sequence of commits will pass locally, and will fail in Gitlab (newest commit on top):

```
abcdef Unrelated code
123456 Code where I break Python formatting
```

In fact, all multi-commit PRs will be vulnerable to passing locally and failing on Gitlab.

The commit range logic is hard to maintain and was only ever meant for the commit checker, where we have no choice. The reason is that the commit checker checks *history*, which cannot be changed after you have merged it. Therefore we need to check only certain ranges. Code however, we can change, therefore I think it should only check what is in the tree, and not care about history at all. I would just remove all the stuff that has to do with Git, and keep all code formatted always.

oleorhagen:
> > Hmm, idea was that a PR should not fail on code that is unrelated to what you have committed yourself.
> 
> It never will, if all code is always checked.

Well, that is true for both scenarios.

> > I don't understand the 'running it locally' critique? Why would it make it harder?
> 
> This sequence of commits will pass locally, and will fail in Gitlab (newest commit on top):
> 
> ```
> abcdef Unrelated code
> 123456 Code where I break Python formatting
> ```
> 
> In fact, all multi-commit PRs will be vulnerable to passing locally and failing on Gitlab.
> 
> The commit range logic is hard to maintain and was only ever meant for the commit checker, where we have no choice. The reason is that the commit checker checks _history_, which cannot be changed after you have merged it. Therefore we need to check only certain ranges. Code however, we can change, therefore I think it should only check what is in the tree, and not care about history at all. I would just remove all the stuff that has to do with Git, and keep all code formatted always.

I still don't understand how this would pass locally? You don't even run this locally, except if you have mendertesting in your path.

But I do get your critique. I don't agree however. Especially on the dropping history part. That should absolutely be a part of it I think, so that every commit is guaranteed to be formatted correctly. 

I guess the difference is you are advocating a repository guard, and I am advocating a repository sentinel approach. 

Anyways, I guess we should include @lluiscampos in this discussion. And leave this hanging until we can figure out what to do here

kacf:
> > > Hmm, idea was that a PR should not fail on code that is unrelated to what you have committed yourself.
> > 
> > 
> > It never will, if all code is always checked.
> 
> Well, that is true for both scenarios.

Wrong. With the current approach, if you add one single line to a file that has not previously passed the format filter, it will now require you to reformat the whole file. Granted, if you reformat every repository now, it's much less likely that this will happen, but dropping the Git requirement *guarantees* that it will never happen, because no file can ever escape the filter.

> I still don't understand how this would pass locally? 

Indeed, there is hidden complexity here, which is exactly what the problem is about. When you run the tests locally, the Git logic has no way of finding out which commit range you want to check, because there is no PR yet, so it just checks [the last one](https://github.com/mendersoftware/mendertesting/blob/master/check_python_code_format#L74-L78). If you run it on Gitlab, there is [additional logic](https://github.com/mendersoftware/mendertesting/blob/master/check_python_code_format#L50-L72) to find out what you want to check, but the code is fragile, and doesn't work anywhere else besides on Gitlab (Travis is different for instance). And we are constantly running into problems with it because of shallow cloning and other unanticipated environmental factors on the test runners. The last one was [today](https://github.com/mendersoftware/deployments-enterprise/pull/51/files#diff-96edf7a6f008de9e928d04e1ae5e12a5R15)!

And that's my point: Why should we need to understand, deal with, and maintain that here? The format check has one simple task, to check the code formatting, and that's it. It should not care about the history. The KISS principle! :-)

For comparison, see how short our Go code checking is, it's a [one liner](https://github.com/mendersoftware/mendertesting/blob/master/.gitlab-ci.yml#L37)! Probably not realistic for the Python checker, but it's 137 lines currently, and can be trimmed *considerably* if we lose the Git stuff.

> You don't even run this locally, except if you have mendertesting in your path.

Of course we should! Having remote-only tests promotes alienation from the tests. All our tests should run locally as well, or at least be runnable.

> But I do get your critique. I don't agree however. Especially on the dropping history part. That should absolutely be a part of it I think, so that every commit is guaranteed to be formatted correctly.

The current method does not guarantee this, both because files you don't change are omitted, and because we know from experience that our Git commit range detection is fragile, and may simply be wrong in some cases. If we drop the history, then every commit is truly guaranteed.

> I guess the difference is you are advocating a repository guard, and I am advocating a repository sentinel approach.

Not sure I fully understood this. You mean strictly enforced policy as opposed to softly enforced for new commits? Either of those options is fine IMHO, but I'm proposing the first option mainly on technical grounds, because it is much less maintenance.

lluiscampos:
@oleorhagen @kacf So what I wanted to avoid when suggesting to check the formatting only on modified files was to do the huge commit updating all files and hence destroying the git blame history.

However, given the experience of @kacf with this matter and acknowledging the complexity/fragility of the script, I am not having an strong point anymore on this and I rather follow his advice. @oleorhagen I am sorry we need to rework this, but I believe now that it will save us trouble in the future.

oleorhagen:
Alright, full repo check it is.

oleorhagen:
PR: https://github.com/mendersoftware/mendertesting/pull/38

oleorhagen:
Updated with the new full repo check

PR #26, opened 2020-02-11
Fix acceptance tests

lluiscampos:
The tests have been broken for a while, but because we were accepting failures in the Pipeline job it went unnoticed. This PR fixes the specific issues and disallow the job to fail from now on (there is a alternative method to skip tests if required).

See individual commits.

### DIFF
--- a/tests/test_package_client.py
+++ b/tests/test_package_client.py
@@ -180,7 +180,7 @@ class TestPackageMenderClientDefaults(PackageMenderClientChecker):
             mender_dist_packages_versions["mender-client"],
         )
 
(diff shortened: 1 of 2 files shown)


tests/test_package_client.py
```diff
@@ -268,7 +269,7 @@ def test_install_configure_start(
             mender_dist_packages_versions["mender-client"],
         )
 
-        self.check_installed_files(setup_tester_ssh_connection, "rasbperrytest")
+        self.check_installed_files(setup_tester_ssh_connection, "raspberrytest")
```
kacf: This one is hard to spot, even when it's pointed out!

PR #27, opened 2020-03-05
MEN-2948: The systemd mender service file is now mender-client  

oleorhagen:
Changelog: None


lluiscampos:
Oh, so there was a PR for this! Good.

Anyway, I think the other change is better as here we will be breaking the build for Mender 2.3.x (2.2.x mender-client).

oleorhagen:
> Oh, so there was a PR for this! Good.
> 
> Anyway, I think the other change is better as here we will be breaking the build for Mender 2.3.x (2.2.x mender-client).

Yeh, I agreed. So I just closed it :)

PR #28, opened 2020-03-17
Update service name and fix acceptance tests

lluiscampos:
Following rename done in master, see MEN-2984

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,8 +35,9 @@ RUN git checkout $MENDER_VERSION
 COPY debian-master debian-master
 COPY debian-2.1.x debian-2.1.x
 
-# And add systemd service file for the modern recipes
-RUN cp support/mender.service debian-master/mender-client.service
+# And add systemd service file for the recipes. It is called "mender" for 2.3.x and before
+RUN cp support/mender-client.service debian-master/mender-client.service || \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -35,8 +35,9 @@ RUN git checkout $MENDER_VERSION
 COPY debian-master debian-master
 COPY debian-2.1.x debian-2.1.x
 
-# And add systemd service file for the modern recipes
-RUN cp support/mender.service debian-master/mender-client.service
+# And add systemd service file for the recipes. It is called "mender" for 2.3.x and before
+RUN cp support/mender-client.service debian-master/mender-client.service || \
```
oleorhagen: Btw, if you want to be fancy:
`cp support/(mender|mender-client).service debian-master/mender-client.service`

lluiscampos: Really? Is that `bash` specific? I haven't seen that before...

oleorhagen: That'd be `bash` yeh. I guess most modern shells should mimic this though.

lluiscampos:
> LGTM :)

The acceptance tests are failing (but they pass locally). I will use this PR to debug a bit and see if I can find a fix.

lluiscampos:
I think the tests should pass now. @oleorhagen please have a look again.

PR #30, opened 2020-05-28
MEN-3601: Add OpenSSL dependency on Debian package

lluiscampos:
Updated the build script and requirements, as well as the debian control
file shipped in the package.
    
Recipe for master differs now latest releases, so branch out these into
their own debian folder.


Dockerfile
```diff
@@ -34,6 +38,8 @@ RUN git checkout $MENDER_VERSION
 # Copy the debian recipe(s)
 COPY debian-master debian-master
 COPY debian-2.1.x debian-2.1.x
+COPY debian-2.2.x debian-2.3.x
+COPY debian-2.3.x debian-2.2.x
```
kacf, line 42: The destination folder is opposite of the source folder. Typo?

lluiscampos: Yes, typo :) Thanks!

Ironically one is a symlink to the other one, so technically doesn't matter. I'll fix.

lluiscampos:
The acceptance tests require fixes (both here and in `mender-test-containers` repo). I am working on that.

lluiscampos:
> The acceptance tests require fixes (both here and in `mender-test-containers` repo). I am working on that.

Fixed now; @kacf please have a look.

lluiscampos:
@kacf I will launch an independent PR on `mender-test-containers` to fix the other repos, but for this one there was no fix required as the deb installation is done in the test code.

PR #31, opened 2020-06-02
Fix systemd service missing issue for 2.3.x/2.2.x recipes

lluiscampos:
Modify the code to better express the logic: the service file is
required for all debian-* directories except for 2.1.x, and on top of
that can be called mender-client.service or mender.service.


lluiscampos:
Run with master [here](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/pipelines/151937269) and with 2.3.0b1-build3 [here](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/pipelines/151937753).

PR #33, opened 2020-08-24
Minor improvements around Docker image build

lluiscampos:
* Add --no-cache to the Docker build command
The image shall fetch Mender upstream on each build, so add an explicit
flag to disable cache. Mainly to help local builds, does not affect CI.

* Parametrize Docker base image
So that eventually we can reuse the same Dockerfile to build packages
for different Debian versions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM debian:stretch
+ARG DEBIAN_VERSION=stretch
+FROM debian:$DEBIAN_VERSION
 
 RUN dpkg --add-architecture armhf && \
     dpkg --add-architecture arm64 && \
--- a/docker-mender-dist-packages
+++ b/docker-mender-dist-packages
@@ -20,7 +20,7 @@ IMAGE_NAME=mender-dist-packages-builder
 
 docker_build_args="--build-arg MENDER_VERSION=${MENDER_VERSION:master}"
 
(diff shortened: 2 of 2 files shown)


mzedel (approved):
👍 I especially like the parametrized image version. While I doubt dependabot would pick it up in other repos, this could be a good approach to have the golang/ node version in a single location...

Since the docker build doesn't take too long here I think the `--no-cache` should be fitting.
Just in case you didn't already come across this suggestion for the "single uncached image layer" problem: https://github.com/moby/moby/issues/1996#issuecomment-550020843 - when the supplied build arg is read from a time source e.g.

lluiscampos:
> Since the docker build doesn't take too long here I think the `--no-cache` should be fitting.
> Just in case you didn't already come across this suggestion for the "single uncached image layer" problem: [moby/moby#1996 (comment)](https://github.com/moby/moby/issues/1996#issuecomment-550020843) - when the supplied build arg is read from a time source e.g.

The problem here is that the ARG will not change (i.e. will still be `master`) but still we need to do the git clone to get latest.

mzedel:
Passing it alone might not work, but to trigger the change in the layer you could prepend relevant layer with `echo "$FOO" && ` and pass `--build-arg FOO=$(date +%s)` - with adjustable date passing to use the cached layer for days/ months... - hackish but _should_ work

PR #34, opened 2020-10-22
QA-214: Publish a 'latest' Mender deb package

oleorhagen:
The two first commits are just refactoring.

The last one is more of a question:

* Should this be published separately, as an option (Like I have done here)?
* Or should it be published alongside every build the script does?

I'm not really sure what is the preferred option here.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -68,9 +68,9 @@ publish:s3:
     - build
   before_script:
     - apt update && apt install -yyq awscli
+    - deb_version=$(cat output/mender-client-deb-version)
   script:
     - echo "Publishing ${MENDER_VERSION} packages to S3"
-    - deb_version=$(cat output/mender-client-deb-version)
     # For master packages, the Debian version is "0.0~git[iso-date].[git-hash]-1" and we make a copy named "master" for GUI to use
     - for arch in amd64 arm64 armhf; do
         aws s3 cp output/mender-client_${deb_version}_${arch}.deb
@@ -86,3 +86,11 @@ publish:s3:
       done
   only:
     - master
+
+publish:s3:latest:
+  extends: publish:s3
+  only:
+    variables:
+      - $MENDER_VERSION =~ /[0-9]+\.[0-9]+\.[0-9]+/
+  before_script:
+    - deb_version=latest
--- a/mender-deb-package
+++ b/mender-deb-package
@@ -39,7 +39,10 @@ build_packages() {
   dch --create -v ${version} --package mender-client "Release ${version}. See online docs for Changelog"
 
   # Native build (amd64)
-  dpkg-buildpackage -us -uc -b
+  dpkg-buildpackage \
+    --unsigned-source \
+    --unsigned-changes \
+    --build=binary
 
   # ARM 32bit build (custom toolchain to support ARMv6)
   CROSS_COMPILE="arm-buildroot-linux-gnueabihf" \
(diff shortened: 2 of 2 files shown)


lluiscampos (changes requested):
Nice refactoring. There is nothing like fresh eyes on old code :)

I think you miss-understood the task. The idea is to have a "latest" package alongside the tags. Meaning:
1) It only gets updated when there is a new tag (never from a master build)
2) When there is a new tag, the exact same build is published twice (with the tag and with "latest")

IOW, I expected changes only in the pipeline, not at all in the build code. See also comments in JIRA ticket or let's have a call together with Kristian to discuss it.

oleorhagen:
> 
> I think you miss-understood the task. The idea is to have a "latest" package alongside the tags. Meaning:
> 

Most definitely confused on this one, so I just gave it a shot. 



> Nice refactoring. There is nothing like fresh eyes on old code :)
> 
> I think you miss-understood the task. The idea is to have a "latest" package alongside the tags. Meaning:
> 
>     1. It only gets updated when there is a new tag (never from a master build)
> 
>     2. When there is a new tag, the exact same build is published twice (with the tag and with "latest")
> 
> 
> IOW, I expected changes only in the pipeline, not at all in the build code. See also comments in JIRA ticket or let's have a call together with Kristian to discuss it.

@lluiscampos see if the latest commit is more appropriate, I'm still not too confident here, but I _think_ this is more like what you had in mind?

.gitlab-ci.yml
```diff
@@ -34,6 +34,22 @@ build:
...
+  only:
+    variables:
+      - $MENDER_VERSION =~ /[0-9]+\.[0-9]+\.[0-9]+/
+  stage: build
+  services:
+    - docker:dind
+  before_script:
+    - apk --update --no-cache add bash sudo
+  script:
+    - MENDER_VERSION=latest ./docker-mender-dist-packages
```
lluiscampos, line 48: This won't work, as the script (I think!) will try to do `git checkout latest`.

lluiscampos: As discussed in person: we don't need this job

.gitlab-ci.yml
```diff
@@ -86,3 +102,23 @@ publish:s3:
...
+      - $MENDER_VERSION =~ /[0-9]+\.[0-9]+\.[0-9]+/
+  stage: publish
+  image: debian:buster
+  dependencies:
+    - build
+  before_script:
+    - apt update && apt install -yyq awscli
+  script:
+    - echo "Publishing latest packages to S3"
+    - deb_version=$(cat output/mender-client-deb-version)
```
lluiscampos, line 118: This is the variable that we need to override to be "latest".

As discussed in person, this will just make a copy of the latest tagged mender client deb package in S3.

.gitlab-ci.yml
```diff
@@ -86,3 +102,23 @@ publish:s3:
...
+    - for arch in amd64 arm64 armhf; do
+        aws s3 cp output/mender-client_${deb_version}_${arch}.deb
+          s3://${S3_BUCKET_NAME}/latest/${S3_BUCKET_SUBPATH}/${arch}/mender-client_${deb_version}_${arch}.deb;
+        aws s3api put-object-acl --acl public-read --bucket ${S3_BUCKET_NAME}
+          --key latest/${S3_BUCKET_SUBPATH}/${arch}/mender-client_${deb_version}_${arch}.deb;
+      done
```
lluiscampos, line 124: To reuse the code, we could move `deb_version=...` to `before_script` in `publish:s3`, and then make this pipeline job to extend `publish:s3`, by only overriding the `before_script`.

lluiscampos (changes requested):
@oleorhagen Getting conceptually closer, yes! See my comments below.

lluiscampos (approved):
Looks good :+1: 

Could you verify once merged? Just manually start a new pipeline with MENDER_VERSION to current latest tag, and both packages should be published.

This verification will actually create our first "latest" package, which in return will allow you to verify the `mender-convert` PR.

oleorhagen:
https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/207585711

oleorhagen:
https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/207598462

oleorhagen:
https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/207653858

PR #36, opened 2020-10-26
fix: Explicitly upload the deb_package to the /latest/ bucket

oleorhagen:
While also giving it the name latest.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,6 +6,8 @@ variables:
   S3_BUCKET_SUBPATH: "dist-packages/debian"
   # Set to true to skip tests (i.e. old Mender client build). See QA-123
   SKIP_TESTS: "false"
+  # Update the 'latest' debian alias to a new client release
+  PUBLISH_LATEST_CLIENT_PACKAGE: "false"
 
 include:
   - project: 'Northern.tech/Mender/mendertesting'
@@ -91,7 +93,12 @@ publish:s3:latest:
   extends: publish:s3
   only:
     variables:
-      - $MENDER_VERSION =~ /[0-9]+\.[0-9]+\.[0-9]+/
-  before_script:
-    - apt update && apt install -yyq awscli
-    - deb_version=latest
+      - $PUBLISH_LATEST_CLIENT_PACKAGE == "true"
+  script:
+    - echo "Publishing latest packages to S3"
+    - for arch in amd64 arm64 armhf; do
+        aws s3 cp output/mender-client_${deb_version}_${arch}.deb
+          s3://${S3_BUCKET_NAME}/latest/${S3_BUCKET_SUBPATH}/${arch}/mender-client_latest_${arch}.deb;
+        aws s3api put-object-acl --acl public-read --bucket ${S3_BUCKET_NAME}
+          --key latest/${S3_BUCKET_SUBPATH}/${arch}/mender-client_latest_${arch}.deb;
+      done


.gitlab-ci.yml
```diff
@@ -92,6 +92,11 @@ publish:s3:latest:
...
-  before_script:
-    - apt update && apt install -yyq awscli
-    - deb_version=latest
+  script:
+    - echo "Publishing latest packages to S3"
+    - for arch in amd64 arm64 armhf; do
+        aws s3 cp output/mender-client_${deb_version}_${arch}.deb
+          s3://${S3_BUCKET_NAME}/latest/${S3_BUCKET_SUBPATH}/${arch}/mender-client_latest_${arch}.deb;
+        aws s3api put-object-acl --acl public-read --bucket ${S3_BUCKET_NAME}
+          --key ${MENDER_VERSION}/${S3_BUCKET_SUBPATH}/${arch}/mender-client_${deb_version}_${arch}.deb;
```
oleorhagen, line 101: I am unsure as to whether or not this should be:
```suggestion
          --key ${MENDER_VERSION}/${S3_BUCKET_SUBPATH}/${arch}/mender-client_latest_${arch}.deb;
```
or something else?

lluiscampos, line 101: The ACL needs to be applied to the same path as the file:

```suggestion
          --key latest/${S3_BUCKET_SUBPATH}/${arch}/mender-client_latest_${arch}.deb;
```

lluiscampos (changes requested):

.gitlab-ci.yml
oleorhagen: https://github.com/mendersoftware/mender-dist-packages/pull/36#discussion_r511958182

lluiscampos (approved):
I think it looks good :crossed_fingers: 

Sorry I didn't catch this previously.

oleorhagen:
No worries. This one took me a while to comprehend.

oleorhagen:
Also, @lluiscampos for a patch release, like I am doing now, this will have to be disabled, or at least the user should be aware of this, so that for this release, this does not wind up pointing to `2.3.1`.

lluiscampos:
> Also, @lluiscampos for a patch release, like I am doing now, this will have to be disabled, or at least the user should be aware of this, so that for this release, this does not wind up pointing to `2.3.1`.

That is an really good catch, @oleorhagen. As we just discussed in person, we need to go back to the drawing table a bit... The design should have been something like:
* `mender` repo figures out if the tag under test is the latest or not. If so, add an special variable to the [mender-dist-packages trigger](https://github.com/mendersoftware/mender/blob/master/.gitlab-ci.yml#L87).
* `mender-dist-packages` pipeline just accepts an extra env variable `PUBLISH_LATEST` or something like that for which to publish the latest package blindly.

Does it make sense?

The code for the actual publish I think is correct in this PR, is only that the `only` part of the job should now use a new env variable.

.gitlab-ci.yml
```diff
@@ -91,7 +91,12 @@ publish:s3:latest:
   extends: publish:s3
   only:
     variables:
-      - $MENDER_VERSION =~ /[0-9]+\.[0-9]+\.[0-9]+/
-  before_script:
-    - apt update && apt install -yyq awscli
-    - deb_version=latest
+      - $PUBLISH_LATEST_CLIENT_PACKAGE == "true"
```
lluiscampos, line 94: Please add the variable (optionally with a doc string) at the beginning of the pipeline, in `variables` section, with default falue `false`. Keep in mind that we might need to trigger it manually some times so it is good to have it at the top.

lluiscampos (changes requested):

oleorhagen:
@lluiscampos should be good

lluiscampos (approved):

PR #38, opened 2020-11-11
Refactor repository to prepare for more packages

lluiscampos:
Namely:
* Move the recipes into a subfolder with the mender-client prefix
* Move the source checkout from the Dockerfile to the runtime script
* Pass arguments to the entrypoint script so that it can be reused


Dockerfile
```diff
@@ -29,30 +29,9 @@ RUN wget -q https://dl.google.com/go/go$GOLANG_VERSION.linux-amd64.tar.gz \
...
-      # It is called "mender" for 2.2.x and earlier
-      cp support/mender-client.service ${debiandir}/mender-client.service ||  \
-      cp support/mender.service ${debiandir}/mender-client.service;           \
-    done
+COPY recipes /recipes
 
 # Prepare the deb-package script
 COPY mender-deb-package /usr/local/bin/
-ENTRYPOINT bash /usr/local/bin/mender-deb-package
+ENTRYPOINT  ["bash", "/usr/local/bin/mender-deb-package"]
```
oleorhagen, line 37:
```suggestion
ENTRYPOINT  /usr/local/bin/mender-deb-package
```
How about we just give the script a hashbang?

mender-deb-package
```diff
@@ -85,7 +116,13 @@ copy_packages_to_output() {
 
 verify_output_directory_exists
 
-extract_client_version
+verify_script_arguments $@
```
oleorhagen, line 119:
```suggestion
verify_script_arguments "$@"
```

oleorhagen, line 119:
```suggestion
verify_script_arguments "$@"
```

oleorhagen (approved):
So much Yes!

mender-deb-package
```diff
@@ -21,25 +21,56 @@ verify_output_directory_exists() {
...
+  VERSION=$3
+}
+
+checkout_repo() {
+  go get -d $REPO_URL
+  cd /root/go/src/$REPO_URL
+  git checkout $VERSION
+}
+
+get_package_version() {
```
kacf, line 39: Would consider using `get_deb_version` to match the variable name it's storing it in.

mender-deb-package
kacf: +1

kacf: +1

Dockerfile
kacf: Yes, it even has one, but it has a typo, missing the `!`

kacf (approved):
Looks good, just minor adjustments.

Dockerfile
lluiscampos: I'll fix the typo, thanks.

With regards to the syntax change. We need to use the exec form (i.e. `[...]`) for then `docker run` be able to pass parameters to the script. See [Docker documentation](https://docs.docker.com/engine/reference/builder/#entrypoint).

PR #39, opened 2020-11-16
MEN-4092/MEN-4135: mender-shell and APT repository

lluiscampos:
```
commit 72a0481a0cd37d7b4e8c90b80eb18e007634f22d
Date:   Fri Nov 13 14:50:16 2020 +0100

    MEN-4092: Build mender-shell debian package
    
    Add recipe to build mender-shell package the same way we build
    mender-client one.
    
    Rename jobs of the pipeline to make it explicit that only mender-client
    is published in the legacy path, mender-shell will only be published in
    the new repository.

commit 7dd97805fa5ecbc006e3656c49e27c67fb023c65
Date:   Sat Nov 14 18:44:33 2020 +0100

    MEN-4135: Publish deb packages to ATP repository
    
    Couple of changes to make everything work from CI/CD:
    * The built deb package targets now experimental (when built from
      master) or stable (when built from tag).
    * The build deb package is signed if a GPG key is present (iow, when
      building from protected branch). GitLab script and docker wrapper
      modified accordingly
    * New job in GitLab pipeline to recreate reprepro configuration, include
      the new packages, and re-publish the new repo indexes.

commit a23cbd67d0f0d6b345fa8baf0c583e0020bacf3b
Date:   Mon Nov 16 11:39:07 2020 +0100

    Refactor Dockerfile and scripts per architecture
    
    A new dependency for mender-cli (and soon mender-client), package
    libglib2.0-dev, is incompatible among other architectures, so we cannot
    use anymore the same image to cross build for all architectures.
    
    Refactor the logic to use three different Docker images for each of the
    target architectures.
```


mender-deb-package
```diff
@@ -61,9 +61,8 @@ prepare_recipe() {
       cp support/mender-client.service debian/mender-client.service ||
       cp support/mender.service debian/mender-client.service
     fi
-  else
-    # Other packages here
-    true
+  elif [ ${DEB_PACKAGE} = "mender-shell" ]; then
```
oleorhagen, line 65: Nitpick:
Quote both maybe?

recipes/mender-shell/debian-master/control
```diff
@@ -0,0 +1,13 @@
+Source: mender-shell
+Maintainer: Mender Team <mender@northern.tech>
+Section: misc
+Priority: optional
+Standards-Version: 3.9.2
+Build-Depends: debhelper (>= 9)
+
+Package: mender-shell
+Conflicts: mender
```
oleorhagen, line 9: Question:
So the debian package is still `mender` ?

recipes/mender-shell/debian-master/mender-shell.dirs
```diff
@@ -0,0 +1,3 @@
+/etc/mender
```
oleorhagen, line 1: Question:
Not `/etc/mender-shell/` ?

recipes/mender-shell/debian-master/rules
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/make -f
+
+%:
+	dh $@
+
+override_dh_auto_build:
+	make CGO_CFLAGS="${CGO_CFLAGS}" CGO_LDFLAGS="${CGO_LDFLAGS}" CC=${CC} \
```
oleorhagen, line 7: Nit:
Inconsistent quoting. Some are, some arent, of the variables expanded.

recipes/mender-shell/debian-master/control
```diff
@@ -0,0 +1,13 @@
...
+Architecture: any
```
oleorhagen, line 10: Question: Is this really true?

oleorhagen (approved):
Not that this is my area of expertise, but LGTM.

recipes/mender-shell/debian-master/mender-shell.dirs
lluiscampos: The conf file is in `/etc/mender`, see https://github.com/mendersoftware/mender-shell/blob/master/Makefile#L57

I think it makes sense to have in the same directory all Mender related configuration

recipes/mender-shell/debian-master/control
lluiscampos: This should not be here, a leftover from mender-client copy paste

But there is correct, as "mender-client" package conflicts with old "mender" package.

recipes/mender-shell/debian-master/control
lluiscampos: Yes. See https://www.debian.org/doc/debian-policy/ch-controlfields.html#architecture

recipes/mender-shell/debian-master/rules
lluiscampos: Thanks. I'll fix here and in the mender-client recipe

mender-deb-package
lluiscampos: Okey, I'll add them.

As the script checks the arguments, there is no way that DEB_PACKAGE can be empty.

lluiscampos:
An important change in the last push:

```
commit 6b982e53ad0cf34dc4cd6b39d321c1bd7fb49d3b
Date:   Thu Nov 19 15:24:59 2020 +0100

    Rework cross-compiling design for armhf (ARMv6)
    
    Nanemly: use raspbian binaries (compiled for ARMv6) for all dependencies
    that we need at build time.
    
    Dockerfile now has several sections that are target arch dependent, for
    the case of armhf extra logic is added to manually download and extract
    deb packages, fake pkg-config, and set the correctly the includes,
    library paths, and actual library list at compile time.
    
    The resulting binaries have been verified on Raspberry Pi Zero W
    hardware (ARMv6).
    
    There is one detail that I was not able to figure out: the build fails
    when not installing Debian packages libmount1 and zlib1g. I might go
    back to this in a later moment.
    
    The other architectures (amd64, arm64) are cross-compiled following "the
    Debian way".
```

Also a minor details wrt versioning:
```
commit bbe11a931b31127dab2bcbff78e4cf3568b1938b
Date:   Thu Nov 19 16:22:33 2020 +0100

    Improve packages semantic versioning scheme
    
    To make it possible to "upgrade" using semantic versions, for example
    from from 2.4.1 (stable) to ~git20201119.1a308e7 (experimental).
```

recipes/mender-shell/debian-master/control
```diff
@@ -0,0 +1,12 @@
...
+Section: misc
+Priority: optional
+Standards-Version: 3.9.2
+Build-Depends: debhelper (>= 9)
+
+Package: mender-shell
+Architecture: any
+Depends: libc6 (>= 2.12), libglib2.0-0 (>=2.50), mender-client (>=2.5.0)
+Description: Mender Shell
+ Mender is an open source over-the-air (OTA) software updater for embedded Linux devices.
```
merlin-northern, line 12: suggested: 'Mender Shell is a Mender add-on allowing to open remote interactive shell terminal sessions'

recipes/mender-shell/debian-master/control
```diff
@@ -0,0 +1,12 @@
+Source: mender-shell
+Maintainer: Mender Team <mender@northern.tech>
+Section: misc
+Priority: optional
+Standards-Version: 3.9.2
+Build-Depends: debhelper (>= 9)
```
merlin-northern, line 6: why we are not specifying the build depends? as in e.g.: libgio dev packages?

merlin-northern (approved):
two comments/questions lgtm

mender-deb-package
```diff
@@ -38,8 +38,18 @@ checkout_repo() {
...
-  # Create a version from VCS. For master, generate something like 0.0~git20191022.dade697
-  DEB_VERSION=$(git describe --tags --exact-match 2>/dev/null || git log -1 --pretty=0.0~git%cd.%h --date format:%Y%m%d)-1
+  # Create a version from Git. For master, generate X.Y.Z~git20191022.dade697-1,
+  # where X.Y.Z is latest tag (not necessarily matching git describe)
+  if git describe --tags --exact-match 2>/dev/null; then
+    DEB_VERSION="$(git describe --tags --exact-match)-1"
+  else
+    DEB_VERSION="$(git tag | egrep '^[0-9]+\.[0-9]+\.[0-9]+$' | sort -rV | head -n1)"
+    if [ -z "${DEB_VERSION}" ]; then
+      DEB_VERSION="0.0"
```
kacf, line 48: Not `0.0.0`?

kacf (approved):
Great work figuring this out! Your frustration will in the end benefit everyone that installs this package on ARM!

All of the stuff I commented on is pretty minor, and/or questionable, so pre-approving!

recipes/mender-shell/debian-master/control
```diff
@@ -0,0 +1,12 @@
+Source: mender-shell
+Maintainer: Mender Team <mender@northern.tech>
+Section: misc
```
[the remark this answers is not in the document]
lluiscampos: @kacf Thanks for bringing this up. I used `misc` same as we currently have for `mender-client` in this repo.

However now I have checked with Debian official packages, and `mender-artifact`, `mender-cli` and `mender-client` use `devel` (ref [here](https://salsa.debian.org/go-team/packages/mender-client/-/blob/master/debian/control#L2)).

I suggest then we change our packages to use `devel` too. Do you agree?

.gitlab-ci.yml
```diff
@@ -31,7 +35,9 @@ build:
   before_script:
     - apk --update --no-cache add bash sudo
   script:
-    - MENDER_VERSION=$MENDER_VERSION ./docker-mender-dist-packages
+    - MENDER_VERSION=$MENDER_VERSION
+      MENDER_SHELL_VERSION=$MENDER_SHELL_VERSION
+      ./docker-mender-dist-packages
```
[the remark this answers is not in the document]
lluiscampos: Thanks. I will clean-up.

mender-deb-package
lluiscampos: Before this PR we were using `0.0`, which is a notation that I saw in our non-versioned packages in Debian repos (mendertesting, log, scopstack), so I just followed the same.

However I run into a test error now (we have an strict regex to check the version) and it will be more convenient to use `0.0.0`, so I will change it :)

mender-deb-package
```diff
@@ -65,15 +65,32 @@ prepare_recipe() {
...
+    "Release ${DEB_VERSION}. See online docs for complete changelog"
 }
 
 build_packages() {
+  # For PRs and other development branches, we don't have GPG key
+  sign_flags=""
+  if [ "$(gpg --list-keys)" == "" ]; then
+    sign_flags="--unsigned-source --unsigned-changes"
+  else
+    key_id=$(gpg --list-keys --with-colons| awk -F: '/^pub:/ { print $5 }')
```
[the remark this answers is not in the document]
lluiscampos: :+1: 

Changed to `gpg --list-secret-keys --with-colons| awk -F: '/^sec:/ { print $5 }'`

recipes/mender-shell/debian-master/control
lluiscampos: @merlin-northern Thanks, I'll change.

recipes/mender-shell/debian-master/control
lluiscampos: Well spotted :+1:

But  we are not using these `Build-Depends` at all in our build system. I'll deal with this when preparing the official Debian recipes.

recipes/mender-shell/debian-master/control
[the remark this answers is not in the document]
lluiscampos: I changed both `mender-client` (masetr) and `mender-shell` to `devel`.

Dockerfile
```diff
@@ -6,21 +6,70 @@ ARG ARCH=amd64
...
 RUN if [ "${ARCH}" = "armhf" ]; then \
         wget -nc -q https://toolchains.bootlin.com/downloads/releases/toolchains/armv6-eabihf/tarballs/armv6-eabihf--glibc--stable-2018.11-1.tar.bz2 \
         && tar -xjf armv6-eabihf--glibc--stable-2018.11-1.tar.bz2 \
         && rm armv6-eabihf--glibc--stable-2018.11-1.tar.bz2; \
     fi
 
+# Get depdendencies from upstream, manually donwloading deb packages, and fake pkg-config
+RUN if [ "${ARCH}" = "armhf" ]; then \
+        set -e; \
+        ln -s /bin/true /usr/bin/pkg-config; \
```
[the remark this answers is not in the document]
lluiscampos: Maybe not convincing enough, but I wrote:
```
# NOTE: pkg-config is used from go build to obtain cflags and libs required to build C code; however
# for armhf we will explicitly pass these flags for go to use our custom sysroot instead of the system
# one. Alternatively pkg-config supports setting a prefix, but that would require patching the source code.
```

recipes/mender-shell/debian-master/control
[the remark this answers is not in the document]
kacf: `mender-artifact` and `mender-cli` are clearly `devel`, since they are used to build and administer software for distribution. But I'm less sure about `mender-client` and `mender-shell`. Perhaps they fit better under the `utils` section? See [this list](https://packages.debian.org/stable/) for an explanation of the sections. They use the long name in the headers, but you can see the short name by hovering over the link.

Dockerfile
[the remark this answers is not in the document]
kacf: :+1:

recipes/mender-shell/debian-master/control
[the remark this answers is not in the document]
lluiscampos: @kacf `utils` seems better than `devel`. I also checked RAUC package (see `apt-cache show rauc`), and they have their updater under `admin` section. Maybe admin is the best?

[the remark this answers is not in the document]
kacf: Yes, that's a good category.

Dockerfile
```diff
@@ -1,26 +1,81 @@
...
+
+# Get depdendencies from upstream, manually donwloading deb packages, and fake pkg-config.
+# NOTE: pkg-config is used from go build to obtain cflags and libs required to build C code; however
+# for armhf we will explicitly pass these flags for go to use our custom sysroot instead of the system
+# one. Alternatively pkg-config supports setting a prefix, but that would require patching the source code.
+RUN if [ "${ARCH}" = "armhf" ]; then \
+        set -e; \
+        ln -s /bin/true /usr/bin/pkg-config; \
+        curl -f http://raspbian.raspberrypi.org/raspbian/dists/stretch/main/binary-armhf/Packages -o Packages 2>/dev/null; \
+        for pkg in \
```
oleorhagen, line 37: Wow

Dockerfile
```diff
@@ -1,26 +1,81 @@
...
+        done; \
+        rm -f Packages data.tar.xz *.deb; \
+    fi
+
+RUN if [ "${ARCH}" = "arm64" ]; then \
+        apt-get install -y gcc-aarch64-linux-gnu; \
+    fi
 
 # Golang environment, for cross-compiling the Mender client
 ARG GOLANG_VERSION=1.11.5
```
oleorhagen, line 81: Question:
Is it on purpose that we're still on this old an version of Go, or have we just not updated in a while?

mender-deb-package
```diff
@@ -53,51 +64,82 @@ prepare_recipe() {
...
+      # ARM 32bit build (custom toolchain to support ARMv6)
+      CROSS_COMPILE="arm-buildroot-linux-gnueabihf" \
+                  CC="$CROSS_COMPILE-gcc" \
+                  PATH="$PATH:${ARMV6_TOOLCHAIN_ROOT}/bin" \
+                  LD_LIBRARY_PATH="$LD_LIBRARY_PATH:${ARMV6_TOOLCHAIN_ROOT}/lib:${ARMV6_TOOLCHAIN_ROOT}/usr/lib" \
+                  CGO_CFLAGS="-I${ARMV6_TOOLCHAIN_ROOT}/usr/include \
+                              -I${ARMV6_TOOLCHAIN_ROOT}/usr/include/arm-linux-gnueabihf \
+                              -pthread \
+                              -I${ARMV6_TOOLCHAIN_ROOT}/usr/include/glib-2.0 \
+                              -I${ARMV6_TOOLCHAIN_ROOT}/usr/lib/arm-linux-gnueabihf/glib-2.0/include/" \
```
oleorhagen, line 120: hah, this looks like fun

oleorhagen (approved):
:100:

.gitlab-ci.yml
```diff
@@ -60,10 +76,48 @@ test:acceptance:
...
+    # Include the build packages in stable or experimental
+    - for change_file in $(ls ${CI_PROJECT_DIR}/output/*.changes); do
+    -   reprepro include stable $change_file ||
+        reprepro include experimental $change_file
+    - done
+    # Upload to bucket
+    - aws s3 sync db s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/db
+    - aws s3 sync pool s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/pool
+    - aws s3 ls --recursive s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/pool/ |
+      awk '{cmd="aws s3api put-object-acl --acl public-read --bucket ${S3_BUCKET_NAME} --key "$4; system(cmd)}'
```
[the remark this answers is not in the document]
lluiscampos: It works, I tested it :). I think that the variable expansion happens inside `awk`

Thanks for your suggestion. As this is tricky part to re-test (requires pipeline hacking), I will merge as-is...

Dockerfile
lluiscampos: Not on purpose, just something that we have not updated in a long while. I will follow-up in a separate PR for this, aligning it with today's yocto `dunfell` golang's version.

PR #40, opened 2020-11-20
Misc: dependabot and golang version

lluiscampos:
* [pipeline] Minor beautification: image definition
* Introduce dependabot for pip and gitmodules
* Update golang to version 1.14.7

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -74,7 +74,7 @@ test:acceptance:
   script:
     - apk --update --no-cache add python3 gcc openssh make openssl-dev libffi-dev libc-dev python3-dev
     - cd tests
-    - pip3 install --upgrade -r requirements.txt
+    - pip3 install -r requirements.txt
     - mender_client_deb_version=$(cat ${CI_PROJECT_DIR}/output/mender-client-deb-version)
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -74,7 +74,7 @@ test:acceptance:
   script:
     - apk --update --no-cache add python3 gcc openssh make openssl-dev libffi-dev libc-dev python3-dev
     - cd tests
-    - pip3 install --upgrade -r requirements.txt
+    - pip3 install -r requirements.txt
```
oleorhagen, line 77: `Introduce dependabot and golang version` and... ?

oleorhagen (approved):
Nice !

lluiscampos (commented):
Note to self: requires manual testing on Raspberry Pi Zero W

.gitlab-ci.yml
lluiscampos: I consider this part of dependabot update; we use fixed requirements hence we don't need the upgrade flag.

lluiscampos: I'll add a comment in the commit message

PR #46, opened 2020-11-23
install-mender.sh: client+shell install script

mchalski:
Issues: MEN-4150

WIP

### DIFF
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -110,17 +110,18 @@ get_deps() {
         apt-transport-https \
         ca-certificates \
         curl \
-        gnupg-agent \
-        software-properties-common
+        gnupg-agent
 }
 
 add_repo() {
     curl -fsSL $REPO_URL/gpg | apt-key add -
 
-    add-apt-repository \
-        "deb [arch=$ARCH] $REPO_URL \
-        $CHANNEL \
-        main"
+    repo="deb [arch=$ARCH] $REPO_URL $CHANNEL main"
+
+    if ! grep -F "$repo" /etc/apt/sources.list; then
+	echo "adding $repo to /etc/apt/sources.list"
+        echo "$repo" >> /etc/apt/sources.list
+    fi
 }
 
 inst_component() {
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -1,0 +1,144 @@
+#!/bin/sh
+
+set -e
+
+CHANNEL="stable"
+
+DEFAULT_COMPONENTS="mender-client mender-shell"
+SELECTED_COMPONENTS=$DEFAULT_COMPONENTS
+
+export DEBIAN_FRONTEND=noninteractive
+
+banner (){
+    echo "
+                          _
+ _ __ ___   ___ _ __   __| | ___ _ __ 
+| '_ \` _ \ / _ \ '_ \ / _\` |/ _ \ '__|
+| | | | | |  __/ | | | (_| |  __/ |   
+|_| |_| |_|\___|_| |_|\__,_|\___|_|   
+
+Running the Mender installation script.
+--
+"
+                                      
+}
+
+usage() {
+    echo "usage: install-mender.sh [-h] [-c channel] [component...]"
+    echo "  -h             print this help"
+    echo "  -c CHANNEL     channel: stable(default)|experimental"
+    echo "  <component>    list of components to install"
+    echo ""
+    echo "Supported components (all installed by default):"
+    for c in $DEFAULT_COMPONENTS
+    do
+        echo "  * $c"
+    done
+}
+
+is_known_component() {
+    for known in $DEFAULT_COMPONENTS
+    do
+        if [ "$1" = "$known" ]; then
+            return 0 
+        fi
+    done
+    return 1
+}
+
+parse_args() {
+    selected_components=""
+    while [ $# -gt 0 ]
+    do
+        case $1 in
+            -h)
+                usage
+                exit 0
+                ;;
+            -c)
+                if [ -n "$2" ]; then
+                    CHANNEL=$2
+                    shift 2
+                else
+                    echo "ERROR: channel requires a non-empty option argument."
+                    echo "Aborting."
+                    exit 1
+                fi
+                ;;
+            *)
+                if is_known_component "$1"; then
+                    selected_components="$selected_components $1 "
+                    shift
+                else
+                    echo "Unsupported argument: \`$1\`"
+                    echo "Run \`mender-install.sh -h\` for help."
+                    echo "Aborting."
+                    exit 1
+                fi
+                ;;
+        esac
+    done
+    if [ -n "$selected_components" ]; then
+        SELECTED_COMPONENTS="$selected_components"
+    fi
+}
+
+print_components() {
+    echo "  Selected components:"
+    for c in $SELECTED_COMPONENTS
+    do
+        printf "\t%s\n" "$c"
+    done
+}
+
+init() {
+    REPO_URL=https://downloads.mender.io/repos/debian
+
+    parse_args "$@"
+
+    ARCH=$(dpkg --print-architecture)
+    echo "  Detected architecture:"
+    printf "\t%s\n" "$ARCH"
+
+    echo "  Installing from channel:"
+    printf "\t%s\n" "$CHANNEL"
+}
+
+get_deps() {
+    apt-get update -qq
+    apt-get install -y -qq --no-install-recommends \
+        apt-transport-https \
+        ca-certificates \
+        curl \
+        gnupg-agent \
+        software-properties-common
+}
+
+add_repo() {
+    curl -fsSL $REPO_URL/gpg | apt-key add -
+
+    add-apt-repository \
+        "deb [arch=$ARCH] $REPO_URL \
+        $CHANNEL \
+        main"
+}
+
+inst_component() {
+    echo "Installing $1"
+}
+
+do_install() {
+    apt-get update 
+    apt-get install -y $SELECTED_COMPONENTS
+
+    echo "  Success!"
+    echo "  Please run \`mender setup\` to configure the client."
+    exit 0
+}
+
+banner
+init "$@"
+print_components
+get_deps
+add_repo
+do_install
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -129,7 +129,10 @@ inst_component() {
 
 do_install() {
     apt-get update 
-    apt-get install -y $SELECTED_COMPONENTS
+    apt-get install -y \
+       -o Dpkg::Options::="--force-confdef" \
+       -o Dpkg::Options::="--force-confold" \
+       $SELECTED_COMPONENTS
 
     echo "  Success!"
     echo "  Please run \`mender setup\` to configure the client."
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -110,7 +110,7 @@ get_deps() {
         apt-transport-https \
         ca-certificates \
         curl \
-        gnupg-agent
+        gnupg
 }
 
 add_repo() {
--- a/tests/test_install_mender_sh.py
+++ b/tests/test_install_mender_sh.py
@@ -1,0 +1,157 @@
+#!/usr/bin/python3
+# Copyright 2020 Northern.tech AS
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


lluiscampos:
@mchalski Looks promising!

As discussed privately, we could have support for the user selecting what to install (by default all, but allow for mender-client only install or even no packages install). This way also we can start using it now for tests, for example, without ugly hacks

(The hacks would be needed because as per today mender-shell is not available at stable (no tag) and not install-able at experimental (depends on un-released mender-client 2.5.0))

I will do my full review later, then.

scripts/install-mender.sh
```diff
@@ -0,0 +1,147 @@
...
+
+export DEBIAN_FRONTEND=noninteractive
+
+banner (){
+    echo "
+                          _
+ _ __ ___   ___ _ __   __| | ___ _ __ 
+| '_ \` _ \ / _ \ '_ \ / _\` |/ _ \ '__|
+| | | | | |  __/ | | | (_| |  __/ |   
+|_| |_| |_|\___|_| |_|\__,_|\___|_|
```
kacf, line 19: Love it!

scripts/install-mender.sh
```diff
@@ -0,0 +1,147 @@
...
+
+usage() {
+    echo "usage: install-mender.sh [-h] [-c channel] [component...]"
+    echo "  -h             print this help"
+    echo "  -c             channel: stable(default)|experimental"
+    echo "  <component>    list of components to install"
+    echo ""
+    echo "Supported components:"
+    echo "  * mender-client (default if none provided)"
+    echo "  * mender-shell"
```
kacf, line 35: [MEN-4150](https://tracker.mender.io/browse/MEN-4150) says both should be installed. Should it be changed?

scripts/install-mender.sh
```diff
@@ -0,0 +1,147 @@
...
+
+is_known_component() {
+    for known in $COMPONENTS
+    do
+	if [ "$1" = "$known" ]	
+ 	then
+	  return 0 
+	fi
+    done
+    return 1
```
kacf, line 46: Generally, in the file, there is a mix of indentation, switching between two and four spaces. Would be nice if it was cleaned up (I think four spaces it what we use in most other files).

scripts/install-mender.sh
```diff
@@ -0,0 +1,147 @@
...
+    echo "Installing $1"
+    apt-get install -y -qq "$1"
+}
+
+do_install() {
+    apt-get update 
+
+    for c in $SELECTED_COMPONENTS
+    do
+        inst_component "$c"
```
kacf, line 134: I think this will not work if the user happens to list the components in opposite order, because mender-shell depends on mender-client, which will not be installed until the second step. It should be fine to just install both in one go, apt supports multiple install targets.

kacf (changes requested):

scripts/install-mender.sh
lluiscampos: By default we should install all (see my yesterday's comment). Then we will be fulfilling the acceptance criteria, while having options to actually test it today (we cannot install both as per today, as I tried to explain also in my comment).

scripts/install-mender.sh
```diff
@@ -0,0 +1,147 @@
+#!/bin/sh
+
+set -e
+
+CHANNEL="stable"
+
+DEFAULT_COMPONENTS="mender-client"
+COMPONENTS="$DEFAULT_COMPONENTS mender-shell"
+SELECTED_COMPONENTS=$DEFAULT_COMPONENTS
```
lluiscampos, line 9: As commented before, by default we should install all.

scripts/install-mender.sh
```diff
@@ -0,0 +1,147 @@
...
+Running the Mender installation script.
+--
+"
+                                      
+}
+
+usage() {
+    echo "usage: install-mender.sh [-h] [-c channel] [component...]"
+    echo "  -h             print this help"
+    echo "  -c             channel: stable(default)|experimental"
```
lluiscampos, line 30: One requires an argument, the other none. This is confusing to me, I suggest:
```suggestion
    echo "  -h                    print this help"
    echo "  -c CHANNEL  channel: stable(default)|experimental"
```

scripts/install-mender.sh
```diff
@@ -0,0 +1,147 @@
...
+                    CHANNEL=$2
+	            shift 2
+		else
+                    echo "ERROR: channel requires a non-empty option argument."
+                    echo "Aborting."
+                    exit 1
+                fi
+		;;
+	    *)
+        	SELECTED_COMPONENTS=""
```
lluiscampos, line 68: Won't this be overridden on every component?

scripts/install-mender.sh
lluiscampos, line 35: I also suggest we generate this list based on a defined variable above to ease maintenance. (As I commented, by default should be all, so that we can print before or after, and then just generate a list based on a variable).

scripts/install-mender.sh
lluiscampos: That is correct.

Also we have to experiment a bit on what happens if `mender-client` is already installed, which may happen 1) with a user that has installed an (assume old) package from Debian upstream or 2) a pre-build image which contains the binary, inventory scripts, etc but was not installed as a deb package (which is the case for onboarding).

For these, adding some force flags should do the job.

lluiscampos: Look at [GUI code](https://github.com/mendersoftware/gui/blob/f84425618f867476c95f401a4125d627220358fa/src/js/helpers.js#L430) for inspiration.

lluiscampos (commented):
I did a full review now; a bunch of comments below.

Let me know if you need help with the test!

scripts/install-mender.sh
mchalski: oh yes, sorry, that was a git push mistake.
now it's `+DEFAULT_COMPONENTS="mender-client mender-shell" `.

scripts/install-mender.sh
mchalski: argh, yeah, sorry - will fix.

it's because in part I'm editing the file in an `rw` mounted docker file :)  but the vim there is rudimentary and has default tab/space expansion settings.

scripts/install-mender.sh
mchalski: I fixed the basic remark here: the list is now really a var, and everything is installed in one go.

I'll play some more with the force flags, these will come later.

kacf (approved):

lluiscampos:
@mchalski Are you planning to add the acceptance test to this PR or will you merge and open a new one?

mchalski:
@lluiscampos I'll add it here if you don't mind

scripts/install-mender.sh
```diff
@@ -0,0 +1,144 @@
...
+    echo "  Success!"
+    echo "  Please run \`mender setup\` to configure the client."
+    exit 0
+}
+
+banner
+init "$@"
+print_components
+get_deps
+add_repo
```
lluiscampos, line 143: It came to my mind now, specifically after reviewing the docs PR, that we should check if the repo is already installed before installing it. See [this comment](https://github.com/mendersoftware/mender-docs/pull/1246#discussion_r531145192).

It wasn't part of the acceptance criteria for the script to be idempotent, but it makes sense to me. Is this something easy to check?

mchalski: oops, I only just noticed your remark, but yes, it should be fairly easy.

tests/test_install_mender_sh.py
```diff
@@ -0,0 +1,153 @@
...
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+import http.server
+import socketserver
+import threading
+import os
+import pytest
+
+SSERVER_PORT = 8000
```
lluiscampos, line 22: Is this a typo?
```suggestion
SERVER_PORT = 8000
```

tests/test_install_mender_sh.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+@pytest.fixture(scope="session")
+def script_server():
+    sdir = os.path.join(os.path.dirname(__file__), "..", "scripts")
+    os.chdir(sdir)
```
lluiscampos, line 28: Won't this possibly affect other tests?

tests/test_install_mender_sh.py
```diff
@@ -0,0 +1,153 @@
...
+            channel = "-c " + channel
+
+        setup_tester_ssh_connection_f.run(
+            "curl http://{}:{}/install-mender.sh | sudo bash -s -- {}".format(
+                localhost, SSERVER_PORT, channel
+            )
+        )
+
+        setup_tester_ssh_connection_f.run("dpkg -s mender-client")
+        setup_tester_ssh_connection_f.run("dpkg -s mender-shell")
```
lluiscampos, line 82: For better readability please use the full option
```suggestion
        setup_tester_ssh_connection_f.run("dpkg --status mender-client")
        setup_tester_ssh_connection_f.run("dpkg --status mender-shell")
```

lluiscampos, line 82: By the way, nice way of checking! I didn't know about this option.

tests/test_install_mender_sh.py
```diff
@@ -0,0 +1,153 @@
...
+        try:
+            setup_tester_ssh_connection_f.run(
+                "curl http://{}:{}/install-mender.sh | bash -s -- unknown".format(
+                    localhost, SSERVER_PORT
+                )
+            )
+        except Exception as ex:
+            assert "Unsupported argument: `unknown`" in str(ex)
+        else:
+            assert False, "command should fail"
```
lluiscampos, line 126: This works, but a cleaner way is using the `warn` parameter of Fabric (obs, this is not available in integration tests but only on the other frameworks)

Something in the lines of:
```suggestion
        setup_tester_ssh_connection_f.run(
            "curl http://{}:{}/install-mender.sh | bash -s -- unknown".format(
                localhost, SSERVER_PORT
            ),
            warn=True
        )
        assert result.exited == 1
        assert "Unsupported argument: `unknown`" in result.stderr
```

tests/test_install_mender_sh.py
```diff
@@ -0,0 +1,153 @@
...
+        if channel != "":
+            channel = "-c " + channel
+
+        setup_tester_ssh_connection_f.run(
+            "curl http://{}:{}/install-mender.sh | sudo bash -s -- {} mender-client".format(
+                localhost, SSERVER_PORT, channel
+            )
+        )
+
+        setup_tester_ssh_connection_f.run("dpkg -s mender-client")
```
lluiscampos, line 105: I would explicitly check here that mender-shell was not installed (up to you to do try/except or my suggested warn=True)

tests/test_install_mender_sh.py
```diff
@@ -0,0 +1,153 @@
...
+
```
lluiscampos, line 106: From here on, this would make more sense to test it as part of the default call (but yeah, it is `skip` at the moment). If you agree with that, we should add a TODO here to clean it up.

tests/test_install_mender_sh.py
```diff
@@ -0,0 +1,153 @@
...
+        result = setup_tester_ssh_connection.run(
+            "ip route | grep default | awk '{print $3}'"
+        )
+        return result.stdout.strip()
+
+    @pytest.mark.usefixtures("setup_test_container_f")
+    @pytest.mark.parametrize("channel", ["", "stable", "experimental"])
+    @pytest.mark.skip(
+        reason="no mender-shell in 'stable'; in 'experimental' depends on client 2.5.0 which is not there yet"
+    )
```
lluiscampos, line 59: FYI, I created a task for myself to enable these tests during release, ref [MEN-4193](https://tracker.mender.io/browse/MEN-4193).

lluiscampos (changes requested):
The tests look really good: well organized and easy to read. See some nitpicks below.

Also, the CI fails with `fixture 'setup_test_container_f' not found`. Just update the submodule as part of this PR and we should be fine with that.

tests/test_install_mender_sh.py
mchalski: it was supposed to mean 'script server', but I guess not too intuitive.
I'll name it SCRIPT_SERVER, so that people won't be confused that we're running the mender server.

tests/test_install_mender_sh.py
mchalski: it will, I'll fix it

mchalski:
@lluiscampos I fixed and resolved your latest remarks.
build is still in progress but PR is ready for re-review.

lluiscampos (approved):
Looks good to me.

PR #48, opened 2020-11-25
APT repository: keep unreferenced files

lluiscampos:
So that we can download user specified versions from mender-convert or
possibly other scripts.


tranchitella (approved):
Well done

PR #52, opened 2020-12-05
MEN-4072: add glib dependency to the mender-client Debian package

tranchitella:
Changelog: title

Signed-off-by: Fabio Tranchitella <fabio@tranchitella.eu>

### DIFF
--- a/recipes/mender-client/debian-master/control
+++ b/recipes/mender-client/debian-master/control
@@ -8,6 +8,6 @@ Build-Depends: debhelper (>= 9)
 Package: mender-client
 Conflicts: mender
 Architecture: any
-Depends: libc6 (>= 2.12), liblzma5 (>= 5.1), libssl1.1 (>= 1.1)
+Depends: libc6 (>= 2.12), libglib2.0-0 (>=2.50), liblzma5 (>= 5.1), libssl1.1 (>= 1.1)
 Description: Mender client
  Mender is an open source over-the-air (OTA) software updater for embedded Linux devices.


recipes/mender-client/debian-master/control
```diff
@@ -8,6 +8,6 @@ Build-Depends: debhelper (>= 9)
 Package: mender-client
 Conflicts: mender
 Architecture: any
-Depends: libc6 (>= 2.12), liblzma5 (>= 5.1), libssl1.1 (>= 1.1)
+Depends: libc6 (>= 2.12), libglib2.0-0 (>=2.50), liblzma5 (>= 5.1), libssl1.1 (>= 1.1)
```
kacf, line 11: Where does the number `2.50` come from?

kacf (approved):

recipes/mender-client/debian-master/control
tranchitella: Same dependency we have in Mender shell:
https://github.com/mendersoftware/mender-dist-packages/blob/master/recipes/mender-shell/debian-master/control#L10

To be honest, I trusted @lluiscampos on this

lluiscampos: We build the deb packages from Debian `stretch`. Current version of this library there is `2.50.3-2+deb9u2`, so I took 2.50 as the dependency for us.

FYI, this should not be done manually, but using debian helpers env variables, as it is done for instance in our [upstream packages](https://salsa.debian.org/go-team/packages/mender-client/-/blob/master/debian/control#L32). The reason we have to specify them by hand is the way we build with custom toolchain for ARM v6. If someone has the time to investigate, I'll be happy to hear alternatives :)

lluiscampos (approved):

PR #53, opened 2020-12-09
mender-shell: Set mender-client dependency to 2.4.2-git version

lluiscampos:
* [Dockerfile] Remove echo of private key
Probably a leftover from development debug.

* Bump bugfix version for master/experimental builds
So that 2.4.2~giti[...] evaluates as newer than 2.4.1 but still lower
than 2.4.2 (or 2.5.0).

* mender-shell: Set mender-client dependency to 2.4.2-git version
So that it can be installed in experimental distribution.
The specific version has been selected as what is mender master today,
so that it includes the recently merged DBus feature and the new
versioning scheme from mender-dist-packages (see 868b6b1).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -89,7 +89,6 @@ COPY recipes /recipes
 
 # Import GPG key, if set
 ARG GPG_KEY_BUILD=""
-RUN echo $GPG_KEY_BUILD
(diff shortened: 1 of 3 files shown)


lluiscampos:
This change should unblock https://github.com/mendersoftware/gui/pull/1111

tranchitella (approved):

PR #54, opened 2020-12-10
[Dockerfile] Download and extract Packages.gz

lluiscampos:
Switch to compressed download.

It seems like the download is timing out in CI, maybe because the
upstream server is having performance issues.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,8 @@ RUN if [ "${ARCH}" = "armhf" ]; then \
 RUN if [ "${ARCH}" = "armhf" ]; then \
         set -e; \
         ln -s /bin/true /usr/bin/pkg-config; \
-        curl -f http://raspbian.raspberrypi.org/raspbian/dists/stretch/main/binary-armhf/Packages -o Packages 2>/dev/null; \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -33,7 +33,8 @@ RUN if [ "${ARCH}" = "armhf" ]; then \
 RUN if [ "${ARCH}" = "armhf" ]; then \
         set -e; \
         ln -s /bin/true /usr/bin/pkg-config; \
-        curl -f http://raspbian.raspberrypi.org/raspbian/dists/stretch/main/binary-armhf/Packages -o Packages 2>/dev/null; \
+        curl -f http://raspbian.raspberrypi.org/raspbian/dists/stretch/main/binary-armhf/Packages.gz -o Packages.gz; \
+        gunzip Packages.gz; \
```
merlin-northern, line 37: I always use `gzip -d` do not ask me why, and feel free to pretend I did not say that.

merlin-northern (approved):
lgtm!

Dockerfile
lluiscampos: It is the same, according to the man page.

PR #56, opened 2020-12-11
[mender-client rules] Add install-dbus target

lluiscampos:
In a way that the recipe becomes future ready for more install-* kind of
targets.

### DIFF
--- a/recipes/mender-client/debian-master/rules
+++ b/recipes/mender-client/debian-master/rules
@@ -7,10 +7,10 @@ override_dh_auto_build:
 	make CGO_CFLAGS="${CGO_CFLAGS}" CGO_LDFLAGS="${CGO_LDFLAGS}" CC="${CC}" \
 	GOOS=linux CGO_ENABLED=1 build
 
+# All install targets except install-systemd
+install_targets = $(shell egrep ^install\-.*: Makefile | grep -v install-systemd | cut -d: -f1)
 override_dh_auto_install:
-	# All install targets except install-systemd
-	DESTDIR=debian/mender-client make install-bin install-conf install-identity-scripts \
-	install-inventory-scripts install-modules install-examples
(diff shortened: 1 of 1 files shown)


recipes/mender-client/debian-master/rules
```diff
@@ -7,10 +7,10 @@ override_dh_auto_build:
 	make CGO_CFLAGS="${CGO_CFLAGS}" CGO_LDFLAGS="${CGO_LDFLAGS}" CC="${CC}" \
 	GOOS=linux CGO_ENABLED=1 build
 
+# All install targets except install-systemd
+install_targets = $(shell egrep ^install\-.*: Makefile | grep -v install-systemd | cut -d: -f1)
```
merlin-northern, line 11: indent?

merlin-northern, line 11: seems hacky -- thats why I like it. what if we decide to add another target in the Makefile called install-systemd-redux? I am not super sure about parsing Makefile in here. but you are the expert.

lluiscampos: I agree that is hacky.

However the current approach has already proven difficult to maintain (we keep forgetting to add targets). Weird targets in the future could break this, but that is not worse than today's situation IMO.

In the other hand, your future example would do no harm, as it will install the `-redux` and every other target, so in effect it will install all (which is what I think we want overall).

lluiscampos: > In the other hand, your future example would do no harm, as it will install the -redux and every other target, so in effect it will install all (which is what I think we want overall).

This kind of happens now, doing `make ... install-inventory-scripts install-inventory-local-scripts install-inventory-network-scripts ...`, which is similar to your `-redux` example.

merlin-northern: ok then.

merlin-northern (approved):

PR #68, opened 2021-01-20
[test_install_mender_sh.py] Enable all tests

lluiscampos:
* [test_install_mender_sh.py] Enable all tests
Now that mender-connect 1.0.0 has been released.

* [test_install_mender_sh.py] De-scope some test cases
Remove the channel parametrization for test_client and test_connect,
reducing the overall test cases for this file from 9 to 5.
The channel switch is already tested in the main test case, and it is
very low risk to remove it from the others.
Currently the full test suite takes more than 1 hour, exceeding the
timeout in CI.

### DIFF
--- a/tests/test_install_mender_sh.py
+++ b/tests/test_install_mender_sh.py
@@ -99,24 +99,17 @@ class TestInstallMenderScript:
         assert "Unsupported argument: `unknown`" in res.stdout
 
     @pytest.mark.usefixtures("setup_test_container_f")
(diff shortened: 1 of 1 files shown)


lluiscampos:
@mchalski I pushed a second commit removing some test cases due to the original design being too slow for GitLab. What do you think?

mchalski:
@lluiscampos good optimization, I don't mind at all! all around LGTM.

mchalski (approved):

PR #69, opened 2021-01-20
[pipeline] Add lock around APT repository update

lluiscampos:
When two pipelines run in parallel (which happens on releases) there is
a high change of corrupting the repository.

Prevent it by locking the S3 bucket that we are about to update.

Note that this is not atomic and there is a chance of forever block one
of the pipelines, in which case it will be killed by GitLab timeout. Not
a very high risk.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -96,6 +96,17 @@ publish:s3:apt-repo:
     - fi
     - echo "$GPG_PRIV_KEY_DIST" | gpg --import
     - echo "$GPG_PUB_KEY_BUILD" | gpg --import
+    # Lock the bucket to block concurrent jobs
+    - while true; do
+    -   if aws s3 ls s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/lock; then
+    -     echo "$S3_BUCKET_REPO_PATH locked, waiting..."
+    -     sleep 10
+    -     continue
+    -   fi
+    -   touch lock
+    -   aws s3 mv lock s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/lock
+    -   break
+    - done
     # Prepare repo from two S3 buckets
     - mkdir repo
     - cd repo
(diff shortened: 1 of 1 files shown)


tranchitella (approved):

.gitlab-ci.yml
```diff
@@ -96,6 +96,17 @@ publish:s3:apt-repo:
...
+    - while true; do
+    -   if aws s3 ls s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/lock; then
+    -     echo "$S3_BUCKET_REPO_PATH locked, waiting..."
+    -     sleep 10
+    -     continue
+    -   fi
+    -   touch lock
+    -   aws s3 mv lock s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/lock
+    -   break
+    - done
```
kacf, line 109: Could be shortened significantly:

```
while aws s3 ls s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/lock; do
    echo "$S3_BUCKET_REPO_PATH locked, waiting..."
    sleep 10
done
touch lock
aws s3 mv lock s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/lock
```

.gitlab-ci.yml
```diff
@@ -128,6 +139,8 @@ publish:s3:apt-repo:
     - aws s3 sync dists s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/dists
     - aws s3 ls --recursive s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/dists/ |
       awk '{cmd="aws s3api put-object-acl --acl public-read --bucket ${S3_BUCKET_NAME} --key "$4; system(cmd)}'
+  after_script:
+    - aws s3 rm s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/lock
```
kacf, line 143: Suggest putting an expiration timer on this file in S3, so it auto deletes itself if it becomes stale. Not sure if it is retained if the file is deleted and recreated though...

.gitlab-ci.yml
lluiscampos: Right. Next time.

.gitlab-ci.yml
lluiscampos: I don't really know how these expiration works. I don't promise, but maybe I rework this later on.

PR #73, opened 2021-02-05
MEN-4435: Build and publish mender-configure Debian package

lluiscampos:
And add test to verify mender-connect and mender-configure deb packages.


docker-mender-dist-packages
```diff
@@ -48,10 +48,21 @@ for arch in amd64 armhf arm64; do
             --volume $(pwd)/output:/output \
             $IMAGE_NAME:cross-${arch} \
             mender-connect \
-            github.com/mendersoftware/mender-connect \
+            https://github.com/mendersoftware/mender-connect \
             ${MENDER_CONNECT_VERSION:-master} \
             ${arch}
 
+    # Build mender-configure deb package (only for native arch)
+    if [ "${arch}" = "amd64" ]; then
```
tranchitella, line 56: Looks good, but I don't get why we build the mender-configure deb package only on amd64.

tranchitella (approved):

docker-mender-dist-packages
lluiscampos: We use `Architecture: all` in control file, meaning that we need to build it only once. This `if` clause here is a bit miss-leading: what it is trying to do is making sure we build it only once on the native architecture (`amd64`). The generated `.deb` package has a suffix `...._all.deb`

Does it make sense?

tranchitella: yes it does, the comment is a bit misleading, though

lluiscampos: Rephrased :+1:

kacf (approved):
Fabio is better than me at reviewing the deb internal stuff, but looks good on the surface.

PR #80, opened 2021-02-15
Add mender-configure to default components and add repo tests.

kacf:

### DIFF
--- a/tests/test_install_mender_sh.py
+++ b/tests/test_install_mender_sh.py
@@ -42,6 +42,14 @@ def script_server():
     print("cleaned up script server")
 
 
+def check_installed(conn, pkg, installed=True):
+    """Check whether the given package is installed on the device given by
+    conn."""
+
+    res = conn.run(f"dpkg --status {pkg}", warn=True)
+    assert (res.return_code == 0) == installed
+
+
 class TestInstallMenderScript:
     def _get_localhost_ip(self, setup_tester_ssh_connection):
         """We need to access the toplevel host's port to curl the script.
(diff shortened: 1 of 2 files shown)


tests/test_install_mender_sh.py
```diff
@@ -56,12 +64,7 @@ def _get_localhost_ip(self, setup_tester_ssh_connection):
     @pytest.mark.usefixtures("setup_test_container_f")
     @pytest.mark.parametrize("channel", ["", "stable", "experimental"])
     def test_default(
-        self,
-        script_server,
-        setup_tester_ssh_connection_f,
-        mender_dist_packages_versions,
-        mender_version,
```
lluiscampos, line 63: Leftovers from a more complex test design. Thanks for cleaning up!

tests/test_install_mender_sh.py
```diff
@@ -56,12 +64,7 @@ def _get_localhost_ip(self, setup_tester_ssh_connection):
     @pytest.mark.usefixtures("setup_test_container_f")
     @pytest.mark.parametrize("channel", ["", "stable", "experimental"])
```
lluiscampos, line 65: My suggestion in the task acceptance criteria is to only test on experimental until the release:
```suggestion
    @pytest.mark.parametrize("channel", ["experimental"])
```

kacf: You're right. I was trying to be smart here by selectively testing each package, but it's tricky because `install-mender.sh` doesn't even execute correctly if `experimental` isn't set. I will just do what you said and skip, which also makes it easier to clean up later.

tests/test_install_mender_sh.py
```diff
@@ -121,17 +128,32 @@ def test_client(
...
+        )
+
+    @pytest.mark.usefixtures("setup_test_container_f")
+    def test_configure(
+        self, script_server, setup_tester_ssh_connection_f,
+    ):
+        localhost = self._get_localhost_ip(setup_tester_ssh_connection_f)
+
+        setup_tester_ssh_connection_f.run(
+            f"curl http://{localhost}:{SCRIPT_SERVER_PORT}/install-mender.sh | sudo bash -s -- -c experimental mender-configure"
```
lluiscampos, line 152: Maybe add a similar `TODO` here to remove `-c experimental` once it is released.

lluiscampos (approved):
Good to go!

PR #84, opened 2021-03-04
Changes needed for mender-configure demos.

kacf:
Best reviewed commit by commit.

### DIFF
--- a/recipes/mender-configure/debian-master/control
+++ b/recipes/mender-configure/debian-master/control
@@ -11,3 +11,17 @@ Depends: dbus, curl, mender-client (>=2.5.0)
 Description: Mender Configure
  Mender Configure is a Mender add-on which enhances the Mender Client providing an Update
  Module to configure the device
+
+Package: mender-configure-timezone
+Architecture: all
+Depends: mender-configure
+Description: Mender Configure Timezone
+ Mender Configure Timezone is an add-on to Mender Configure which enables configuration
+ of timezone on the device
+
+Package: mender-configure-demo
+Architecture: all
+Depends: mender-configure, mender-configure-timezone
+Description: Mender Configure Demo
+ Mender Configure Demo is an add-on to Mender Configure which enables some demo
+ configurations on the device
--- a/recipes/mender-configure/debian-master/rules
+++ b/recipes/mender-configure/debian-master/rules
@@ -4,7 +4,7 @@
 	dh $@
 
 override_dh_auto_install:
-	DESTDIR=debian/mender-configure make install
+	DESTDIR=debian/tmp make install install-scripts install-demo
 
 override_dh_auto_test:
 	true
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -4,7 +4,18 @@ set -e
 
 CHANNEL="stable"
 
-DEFAULT_COMPONENTS="mender-client mender-configure mender-connect"
+AVAILABLE_COMPONENTS="mender-client \
+mender-configure \
+mender-configure-demo \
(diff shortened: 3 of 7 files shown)


recipes/mender-configure/debian-master/control
```diff
@@ -11,3 +11,17 @@ Depends: dbus, curl, mender-client (>=2.5.0)
...
+Package: mender-configure-timezone
+Architecture: all
+Depends: mender-configure
+Description: Mender Configure Timezone
+ Mender Configure Timezone is an add-on to Mender Configure which enables configuration
+ of timezone on the device
+
+Package: mender-configure-demo
+Architecture: all
+Depends: mender-configure, mender-configure-timezone
```
lluiscampos, line 24: Does it really depend on `mender-configure-timezone`? I don't think so

recipes/mender-configure/debian-master/rules
```diff
@@ -4,7 +4,7 @@
 	dh $@
 
 override_dh_auto_install:
-	DESTDIR=debian/mender-configure make install
+	DESTDIR=debian/tmp make install install-scripts install-demo
```
lluiscampos, line 7: Oh, so is this "the right way" of doing it? Are the files picked up from there and into their respectively packages folders?

scripts/install-mender.sh
```diff
@@ -27,12 +38,18 @@ usage() {
...
+    echo "Supported components (x = installed by default):"
+    for c in $AVAILABLE_COMPONENTS
     do
-        echo "  * $c"
+        if echo "$DEFAULT_COMPONENTS" | egrep -q "(^| )$c( |\$)"; then
+            echo -n " (x) "
+        else
+            echo -n " (-) "
+        fi
+        echo "$c"
```
lluiscampos, line 52: Really cool :+1:

lluiscampos (changes requested):
Only the dependency looks incorrect to me. Looks good otherwise, good job!

recipes/mender-configure/debian-master/rules
kacf: Yes, it's a bit strange, but `tmp` is actually documented [here](https://www.debian.org/doc/manuals/maint-guide/dother.en.html#install), as the place to install things. And then the builder will pick it up from there.

recipes/mender-configure/debian-master/control
kacf: The UI will refer to broken functionality unless you install both. This was the thinking behind adding the dependency, to avoid user confusion. Do you think it's a good enough reason, or should we be more "technically correct" and remove it?

recipes/mender-configure/debian-master/rules
lluiscampos: Thanks

recipes/mender-configure/debian-master/control
lluiscampos: Mmm, I think we should make it technically correct. Where you say "the UI" is really "the UI after onboarding", and in that case we have installed both via get.mender.io without this hacky dependency.

kacf: I think it's not that hacky. The demo is the `mender-demo-raspberrypi-led` and the `timezone` scripts. The only reason that `timezone` is its own package is that it is useful outside the demo as well, otherwise it would have been inside the demo package. In that sense I guess the demo package could be considered more of a meta package, like `gnome`.

lluiscampos: okey. Makes sense

lluiscampos (approved):

PR #87, opened 2021-03-29
Work around when master is exactly on a tag name.

kacf:
If we have requested to build master in the build parameters, we want
a SHA-based package version, even if it matches the tag exactly. This
is also expected by our tests, which fail without this fix.

However, the reported version by the tool should be the tag name, if
it's possible. So testing for this when building master is wrong,
because it can be both a tag name and a SHA, depending on what is
built.

### DIFF
--- a/mender-deb-package
+++ b/mender-deb-package
@@ -40,7 +40,7 @@ checkout_repo() {
 get_deb_version() {
   # Create a version from Git. For master, generate X.Y.Z~git20191022.dade697-1,
   # where X.Y.Z is latest tag (not necessarily matching git describe)
-  if git describe --tags --exact-match 2>/dev/null; then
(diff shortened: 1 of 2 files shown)


oleorhagen (approved):

lluiscampos:
This fixes [QA-276](https://tracker.mender.io/browse/QA-263), right?

lluiscampos (approved):

kacf:
Correct, I will include the ticket reference before I merge.

kacf:
@lluiscampos: I had to change even more stuff here, does it make sense to remove that part?

kacf:
This is blocking release progress, and it's only a test which has changed since the last review, so I will merge this, but please take a look if you agree afterwards.

kacf:
> Correct, I will include the ticket reference before I merge.

Bah, I forgot this. But the ticket is updated.

PR #103, opened 2021-05-27
MEN-4105: new pkg mender-client-dev to distribute the D-Bus interface files

tranchitella:
Changelog: title


tests/test_dist_packages_versions.py
```diff
@@ -37,6 +37,9 @@ def test_versions(
     mender_configure_version,
     mender_dist_packages_versions,
 ):
+    verify_package_version(
+        mender_version, mender_dist_packages_versions["mender-client"]
+    )
     verify_package_version(
         mender_version, mender_dist_packages_versions["mender-client"]
     )
```
lluiscampos, line 45: This new test is not adding any value. I don't think it is important to explicitly verify the `-dev` package and we can drop it. Similarly mender-configure generates more than one package and we only verify the version of the base one.

lluiscampos (approved):
Looks good, thank you.

At some point we should stop adding these kind of small low risk tests on the emulated Raspbian target (which is quite slow) and just rely on host inspection of the packages. But that will be next time :smiley:

PR #106, opened 2021-06-04
Fix: Overwrite the existing mender.list file, instead of appending

oleorhagen:
Changelog: None

### DIFF
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -149,10 +149,7 @@ add_repo() {
         fi
     fi
 
-    if ! grep -F "$repo" /etc/apt/sources.list.d/mender.list; then
-        echo "adding $repo to /etc/apt/sources.list.d/mender.list"
-        echo "$repo" >> /etc/apt/sources.list.d/mender.list
-    fi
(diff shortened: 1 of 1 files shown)


alfrunes (approved):

scripts/install-mender.sh
```diff
@@ -151,7 +151,7 @@ add_repo() {
 
     if ! grep -F "$repo" /etc/apt/sources.list.d/mender.list; then
         echo "adding $repo to /etc/apt/sources.list.d/mender.list"
-        echo "$repo" >> /etc/apt/sources.list.d/mender.list
+        echo "$repo" > /etc/apt/sources.list.d/mender.list
     fi
 }
```
lluiscampos, line 156: I meant to substitute this whole `if` block with only: `echo "$repo" > /etc/apt/sources.list.d/mender.list`.

lluiscampos: (for some reason GitHub does not let me put this as a "suggestion")

lluiscampos (changes requested):

lluiscampos (approved):

PR #107, opened 2021-06-19
Fix APT repo update for mender-client-dev package

lluiscampos:
* [pipeline] Do not accept failures on publish experimental packages
    
    All packages are unique since fix in f742f1c, so this workaround is not
    needed anymore.
    
    Unfortunately, it has actually been hiding another issue.

* dpkg-buildpackage: Stop using --build=binary and let the caller specify
    
    According to its man page, --build=binary is alias for --build=any,all
    which resulted in the three `mender-client_..._<arch>.changes` files
    containing the checksum for mender-client-dev and eventually making
    reprepro fail when trying to upload the same package (-dev) twice!
    
    Use instead "any" on the for loop for the architectures, and explicitly
    pass "all" for the platform independent packages. This creates a
    dedicated change files (`..._all.changes`) for the mender-client with only
    the mender-client-dev checksum.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -146,10 +146,10 @@ publish:s3:apt-repo:
     -     reprepro --keepunreferencedfiles include stable $change_file
     -   done
     - fi
-    # Include everything else to experimental. Allow failures to ignore wrong
-    # distribution (final tags) or to ignore checksum mismatches (master rebuilds)
+    # Include everything else to experimental.
+    # Packages names are unique: we append build number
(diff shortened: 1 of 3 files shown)


lluiscampos:
An example of the current failure in master can be seen here:
* [artifacts form build stage](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/jobs/1359437944/artifacts/browse/output/) (see mender-client `.changes` files)
* [error in publish stage](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/jobs/1359438164)

tranchitella (approved):
Looks good

PR #108, opened 2021-06-21
MEN-4591: Workaround to manually upload deb packages for older releases

lluiscampos:


tranchitella (approved):

.gitlab-ci.yml
```diff
@@ -130,21 +130,95 @@ publish:s3:apt-repo:
...
+        current_version=$(reprepro --architecture amd64 --list-format '${version}' list stable ${package_name})
+        if [ -z "${current_version}" ]; then
+          # Package doesn't exist, so we assume is "newer"
+          return 0
+        fi
+        current_version_semver=$(echo $current_version | cut -d"-" -f1 )
+
+        vercomp $package_version $current_version_semver
+        if [ $? -eq 1 ]; then
+          return 0
```
kacf, line 189: Just double checking: It still works if you use reprepro to upload a package with the same version as the newest, right?

kacf (approved):

.gitlab-ci.yml
kacf: Oops, no, I misread this, it's the opposite of what I said, which will of course work. Never mind.

lluiscampos:
Thanks for your reviews.

I realized however that this won't be enough, as the deb package for the release candidates (`1.2.3-build4` kind of tag) go to `experimental` and there I don't have the workaround. Reworking a bit...

lluiscampos:
> I realized however that this won't be enough, as the deb package for the release candidates (`1.2.3-build4` kind of tag) go to `experimental` and there I don't have the workaround. Reworking a bit...

I reworked it now.

I have tested the logic locally, but I would appreciate your eyes on it again.

tranchitella (approved):

PR #109, opened 2021-06-24
APT repo: do not keep unreferenced files for non-tags

lluiscampos:
The pool is more than 1Gb as per today, and all these master packages
are virtually inaccessible - not linked from mender-docs, not used by
mender-convert.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -137,6 +137,13 @@ publish:s3:apt-repo:
         echo "${version}" | egrep -q '^[0-9]+\.[0-9]+\.[0-9]+$'
         return $?
       }
+    # Bash function to check if the string is a build tag
+    - |
+      function is_build_tag () {
+        version=$1
+        echo "${version}" | egrep -q '^[0-9]+\.[0-9]+\.[0-9]+-build[0-9]+$'
+        return $?
+      }
     # Bash function to compare two semantic versions, returns: 0 (=), 1(>) or 2(<)
     #  * "master" always evaluates to greater than any other version
     #  * "build" versions like 1.3.4-build for are evaluated as 1.2.3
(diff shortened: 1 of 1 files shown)


lluiscampos:
This is a RFC at this point, what do you guys think?

kacf:
But unreferenced files, does that also mean old releases?

lluiscampos:
Let me elaborate a bit more here (and also add some reasoning I had yesterday after work that this PR has an unexpected side effect).

The simplest way of a APT repo to work, and the default when using `reprerpo` is to delete files not longer referenced (or, in other words, inaccessible via `apt-get`). We use the `--keepunreferencedfiles` to keep old packages with tags, which are the ones that `mender-convert` could use (user can select between "latest" or an specific _tagged_ version).

If you manually inspect [Packages file for experimental](https://downloads.mender.io/repos/debian/dists/experimental/main/binary-armhf/Packages) you see that only the latest is referred there (at the time of writing for example `pool/main/m/mender-client/mender-client_2.6.1~git20210617.373883a-1+b326624935_armhf.deb`), making every other master package in the pool inaccessible, in the sense that neither APT repo will serve it neither `mender-convert` can guess the name (as the have these weird gitsha-buildnum versions).

So, what this work is trying to achieve, is for the pool to only hold tagged versions (including old tags) and latest master packages

(I hope it makes sense until this point!)

However this **won't work perfectly as it is**. In experimental we have both untagged packages (master) and tagged ones (release candidates). As the code is now, all tagged versions will keep the old one (so say that tag 1.2.3-builld4 keeps 1.2.3~gitSomething) but the moment we finish the release, the next master package will remove the latest build tag (so say 1.2.4~gitSomething is now removing 1.2.3-build4).

This might be a side effect we can leave with.

Alternatively, I was thinking on having a third [hidden, not linked from the docs] distribution only for release candidates. So what we could have `stable` (public, final releases), `experimental` (public, edge software) and `beta` (private, only to be used while release testing).

PR #112, opened 2021-07-12
Several fixes for mender-client < 3.0.0.

kacf:


kacf:
Note that this breaks the test for these files, but there is [a separate task](https://tracker.mender.io/browse/QA-123) for making the tests version aware.

tranchitella (approved):
Looks good, but if I'm not wrong it will build the mender-client-dev package multiple times.

kacf:
It will, but they are just the same, so it has no ill effects.

PR #114, opened 2021-07-14
Fix incorrect source path when copying not-latest packages.

kacf:


kacf:
Cancelling pipeline. Since this is the publish stage, it won't run until it is in master.

tranchitella (approved):

kacf:
Thanks!

PR #115, opened 2021-07-16
Upgrade the Dockerfile to Debian buster

tranchitella:
We require libssl1.1 >= 1.1.1 to enable ECC keys in the Mender Client.
This version of openssl is not available in stretch, thus we upgrade to
buster.

Changelog: none

Requires: https://github.com/mendersoftware/mender-test-containers/pull/72

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,7 +61,7 @@ RUN if [ "${ARCH}" = "armhf" ]; then \
             uuid-runtime \
             uuid-dev \
         ; do \
-            deb_package_url=$(grep Filename Packages | grep ${pkg}_ | grep armhf | tail -n1 | sed 's/Filename: //'); \
+            deb_package_url=$(grep Filename Packages | grep /${pkg}_ | grep armhf | tail -n1 | sed 's/Filename: //'); \
             echo "Downloading ${pkg}..."; \
             filename=$(basename $deb_package_url); \
             curl -L http://raspbian.raspberrypi.org/raspbian/${deb_package_url} -o $filename 2>/dev/null; \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
-ARG DEBIAN_VERSION=stretch
+ARG DEBIAN_VERSION=buster
 FROM debian:$DEBIAN_VERSION
+ARG DEBIAN_VERSION
 
 RUN apt-get update && \
     apt-get install -y \
@@ -35,7 +36,7 @@ RUN if [ "${ARCH}" = "armhf" ]; then \
 RUN if [ "${ARCH}" = "armhf" ]; then \
         set -e; \
         ln -s /bin/true /usr/bin/pkg-config; \
-        curl -f http://raspbian.raspberrypi.org/raspbian/dists/stretch/main/binary-armhf/Packages.gz -o Packages.gz; \
+        curl -f http://raspbian.raspberrypi.org/raspbian/dists/${DEBIAN_VERSION}/main/binary-armhf/Packages.gz -o Packages.gz; \
         gunzip Packages.gz; \
         for pkg in \
             liblzma5 \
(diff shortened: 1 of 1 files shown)


mzedel (approved):

alfrunes (approved):

Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG DEBIAN_VERSION=stretch
+ARG DEBIAN_VERSION=buster
 FROM debian:$DEBIAN_VERSION
```
mzedel, line 3:
```suggestion
ARG DEBIAN_VERSION=buster
FROM debian:$DEBIAN_VERSION
ARG DEBIAN_VERSION
```

based on [this](https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact)

lluiscampos:
We have been deliberately using Debian `stretch` and the older toolchain with ARMv6 to support a wider range of OS and devices, namely Raspberry Pi Zero W. See also this [old discussion](https://github.com/mendersoftware/mender-dist-packages/pull/13#discussion_r328451238).

At the time was only liblzma, today is also libssl and libglib2.0 (for D-Bus) so it might be time to revisit this decision.

I need to think a bit more to form a proper opinion. I'll come back to this later today

kacf:
Following the code that @alfrunes posted in the chat:

```c
#if OPENSSL_VERSION_NUMBER >= 0x1010100fL
...

int X_EVP_DigestSignInit(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
		const EVP_MD *type, ENGINE *e, EVP_PKEY *pkey){
	return EVP_DigestSignInit(ctx, pctx, type, e, pkey);
}
...
#else
...
int X_EVP_DigestSignInit(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
		const EVP_MD *type, ENGINE *e, EVP_PKEY *pkey){
	return 0;
}
...
#endif
```

We could perhaps use `dlopen("EVP_DigestSignInit")` instead?

lluiscampos:
I think we should keep the criteria of building with the previous-to-stable release. However, [next Debian release is coming this month](https://en.wikipedia.org/wiki/Debian_version_history#Debian_11_(Bullseye)), so we can update this soon. But we need a JIRA task to, among other things, test the updated packages on Raspberry Pi Zero W.

I think @kacf suggestion is good regardless: it will make the software in effect check the openSSL version installed in runtime, not at compile time.

lluiscampos:
I think that you think that by updating the submodule you are testing with the new raspbian testing image, but that is not true. The tests use the image from upstream (from Docker Hub). So that part needs manual testing at least once.

tranchitella:
@lluiscampos I'm fully aware updating the submodule doesn't influence the test, and I launched this PR for discussing with you who are experts in these topics.

lluiscampos:
@tranchitella Okey, sorry.

I'll run the tests locally.

tranchitella:
@kacf can we merge this?

alfrunes:
> Following the code that @alfrunes posted in the chat:
> 
> ```c
> #if OPENSSL_VERSION_NUMBER >= 0x1010100fL
> ...
> 
> int X_EVP_DigestSignInit(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
> 		const EVP_MD *type, ENGINE *e, EVP_PKEY *pkey){
> 	return EVP_DigestSignInit(ctx, pctx, type, e, pkey);
> }
> ...
> #else
> ...
> int X_EVP_DigestSignInit(EVP_MD_CTX *ctx, EVP_PKEY_CTX **pctx,
> 		const EVP_MD *type, ENGINE *e, EVP_PKEY *pkey){
> 	return 0;
> }
> ...
> #endif
> ```
> 
> We could perhaps use `dlopen("EVP_DigestSignInit")` instead?

I think that would be tricky as the respective `EVP_` symbols [exist](https://www.openssl.org/docs/man1.0.2/man3/EVP_DigestSignInit.html) for `1.0` as well. Maybe there's some other symbol we could check for?
Anyways, I think we can merge this PR as it's somewhat unrelated to the above issue.

kacf:
Yeah, we probably don't need that fix if we're building for a release that have all of them anyway. I'm ok with merging as is.

PR #118, opened 2021-08-26
MEN-4760: Add Mender Monitor debian package support

oleorhagen:
Changelog: None


.gitlab-ci.yml
```diff
@@ -243,6 +246,20 @@ publish:s3:apt-repo:
...
+    -   if is_final_tag ${MENDER_MONITOR_VERSION}; then
+    -     codename=stable
+    -   else
+    -     codename=experimental
+    -   fi
+    -   for change_file in $(ls ${CI_PROJECT_DIR}/output/mender-monitor_*.changes); do
+    -     reprepro --keepunreferencedfiles include ${codename} ${change_file}
+    -   done
+    - else
+    -   cp ${CI_PROJECT_DIR}/output/mender-monitor*.deb pool/main/m/mender-monitor/
```
oleorhagen, line 260: Not too familiar with the s3 side of this, is this really the right place to put it :thinking: ?

docker-mender-dist-packages
```diff
@@ -78,3 +78,16 @@ docker run --rm \
         ${MENDER_CONFIGURE_VERSION:-master} \
         amd64 \
         ${CI_PIPELINE_ID:-LOCAL}
+
+# mender-monitor
+if [ -n "${MENDER_TEST_BOT_ACCESS_TOKEN}" ]; then
```
oleorhagen, line 83: Custom token added only to the `test-monitor-debian` branch in GitLab for now, but this (I believe), will be required for the protected branches henceforth.

oleorhagen:
Also `Changelogs` seems like it's only used occasionally here, leaving it empty

tests/test_package_addons.py
```diff
@@ -115,3 +115,42 @@ def test_mender_configure(
         setup_tester_ssh_connection.run(
             "test -x /usr/share/mender/inventory/mender-inventory-mender-configure"
         )
+
+    def test_mender_monitor(
```
oleorhagen, line 119: This will have to be guarded, either here, or in the gitlab.ci file, since it's closed source.
I'm not sure what would be nicest here.

.gitlab-ci.yml
lluiscampos: Discussed in person - No, this package needs to go into a different bucket `mender-binaries`.

recipes/mender-monitor/debian-master/control
```diff
@@ -0,0 +1,15 @@
+Source: mender-monitor
+Maintainer: Mender Team <mender@northern.tech>
+Section: admin
+Priority: optional
+Standards-Version: 3.9.2
+Build-Depends: debhelper (>= 9)
+
+Package: mender-monitor
+Architecture: all
+Depends: mender-client (>=3.0) ${misc:Depends}
```
lluiscampos, line 10: It depends on mender-client 2.5 (requires only auth D-Bus API) and `lmdb` (also bash? :thinking: ) See [dependencies from yocto](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-commercial/recipes-mender/mender-monitor/mender-monitor.inc#L6).

docker-mender-dist-packages
```diff
@@ -78,3 +78,16 @@ docker run --rm \
...
+    docker run --rm \
+        --volume $(pwd)/output:/output \
+        $IMAGE_NAME:cross-amd64 \
+        mender-monitor \
+        all \
+        https://mender-test-bot:${MENDER_TEST_BOT_ACCESS_TOKEN}@github.com/mendersoftware/monitor-client \
+        ${MENDER_MONITOR_VERSION:-master} \
+        amd64 \
+        ${CI_PIPELINE_ID:-LOCAL}
```
lluiscampos, line 92: To be consistent with the existing indentation:
```suggestion
    docker run --rm \
            --volume $(pwd)/output:/output \
            $IMAGE_NAME:cross-amd64 \
            mender-monitor \
            all \
            https://mender-test-bot:${MENDER_TEST_BOT_ACCESS_TOKEN}@github.com/mendersoftware/monitor-client \
            ${MENDER_MONITOR_VERSION:-master} \
            amd64 \
            ${CI_PIPELINE_ID:-LOCAL}
```

.gitlab-ci.yml
```diff
@@ -98,6 +99,8 @@ test:acceptance:
       --mender-connect-deb-version $(cat ${CI_PROJECT_DIR}/output/mender-connect-deb-version)
       --mender-configure-version $MENDER_CONFIGURE_VERSION
       --mender-configure-deb-version $(cat ${CI_PROJECT_DIR}/output/mender-configure-deb-version)
+      --mender-monitor-version $MENDER_MONITOR_VERSION
+      --mender-monitor-deb-version $(cat ${CI_PROJECT_DIR}/output/mender-monitor-deb-version)
```
lluiscampos, line 103: (you probably figured this out yourself, but you need to add new these new arguments to `conftest.py`)

docker-mender-dist-packages
lluiscampos, line 83: I have a suggestion for this but be aware that it might break your bash magic for `REPO_URL_CLEANED`.

I generally don't like having hardcoded the user (mender-test-bot) in this script. Basically we are encouraging a developer to export bot credentials to his/her console in order to debug this package. Instead both the user and the token should be bash variables so that we encourage a developer to export his/her own credentials (creating a temporary token) for debugging.

That being said, this could be something like:
```suggestion
if [ -n "${MENDER_PRIVATE_REPO_ACCESS_USER}" -a -n "${MENDER_PRIVATE_REPO_ACCESS_TOKEN}" ]; then
```

Having in the gitlab-ci.yml file `MENDER_PRIVATE_REPO_ACCESS_USER` set to `mender-test-bot`, for example

lluiscampos: Yes, that is correct and is a prize we need to pay in the name of security.

The missing detail then is that the new acceptance tests will only be able to run on protected branches, so we need a decorator to skip it on PRs. (Still a bit rusty from my vacation, so I can't think about a good example for this, but we do have in other test suites...)

tests/test_package_addons.py
```diff
@@ -115,3 +115,42 @@ def test_mender_configure(
...
+            + mender_dist_packages_versions["mender-monitor"]
+            + ")"
+            in result.stdout
+        )
+
+        # Check mender-monitor files
+        setup_tester_ssh_connection.run(
+            "test -x /usr/share/mender-monitor/mender-monitord"
+        )
+        setup_tester_ssh_connection.run("test -x /etc/mender-monitor/monitor.d/log.sh")
```
lluiscampos, line 156: The test is "good enough" in my opinion. These tests are very slow so I am hesitant to test more than the minimum.

lluiscampos (changes requested):
See a bunch of comments below. Generally is looking good :+1: 

I still want to do couple of local tests to give this the green light after your rework.

docker-mender-dist-packages
oleorhagen: Yes, good idea!

tests/test_package_addons.py
oleorhagen: resolved here: https://github.com/mendersoftware/mender-dist-packages/pull/118#discussion_r698292426

lluiscampos: Sorry I missed you original comment here.

lluiscampos:
> Also `Changelogs` seems like it's only used occasionally here, leaving it empty

This repository, by itself, it not "releasable", hence `Changelogs` are not required.

.gitlab-ci.yml
```diff
@@ -98,6 +101,9 @@ test:acceptance:
...
+      --commercial-tests $(echo "${MENDER_PRIVATE_REPO_ACCESS_TOKEN}")
```
lluiscampos, line 106: Please don't - let's find a better way. You could check if the variable is not empty, then set another variable to true. Read [LFF from 17-02-2021](https://script.google.com/a/northern.tech/macros/s/AKfycbxgcRZg4TNHuCwoLJNBK6FWQR3qfUBSNDVViBDTf7ebM4vsSRow/exec) (Search for that date)

.gitlab-ci.yml
```diff
@@ -324,3 +330,15 @@ publish:production:s3:scripts:install-mender-sh:
   only:
     refs:
       - production
+
+publish:s3:mender-monitor:
+  only:
+    - /^(master|[0-9]+\.[0-9]+\.x)$/
```
lluiscampos, line 336: This doesn't make sense in this context, as `mender-dist-packages` repository does not have release branches. We can only publish from master - the version information is given by `MENDER_MONITOR_VERSION`, so it should be fine

lluiscampos (changes requested):

.gitlab-ci.yml
oleorhagen: Lol, yes. of course :see_no_evil:

.gitlab-ci.yml
```diff
@@ -98,6 +101,9 @@ test:acceptance:
       --mender-connect-deb-version $(cat ${CI_PROJECT_DIR}/output/mender-connect-deb-version)
       --mender-configure-version $MENDER_CONFIGURE_VERSION
       --mender-configure-deb-version $(cat ${CI_PROJECT_DIR}/output/mender-configure-deb-version)
+      --mender-monitor-version $MENDER_MONITOR_VERSION
+      --mender-monitor-deb-version $(cat ${CI_PROJECT_DIR}/output/mender-monitor-deb-version)
+      --commercial-tests
```
oleorhagen, line 106: Just makes it easier to separate later

mender-deb-package
```diff
@@ -33,9 +33,20 @@ verify_script_arguments() {
   BUILD_ID=$6
 }
 
+# Clean the URL of the login information
+checkout_repo_clean_local_path() {
```
lluiscampos, line 37: :+1:

lluiscampos (approved):

oleorhagen:
Waiting for: https://github.com/mendersoftware/mender-dist-packages/pull/120

Running on top of this now, rebasing on master after it's merged

PR #120, opened 2021-08-31
test_install_mender_sh.py: Fix apt-get error after suite rename

lluiscampos:
Upstream raspbian.raspberrypi.org has changed the suite name for buster
from "stable" to "oldstable", so the test client needs to explicitly
accept the change before running other apt-get commands.

Moved the script_server fixture to a class level marker and removed the
redundant setup_tester_ssh_connection_f fixture marker (explicit in the
method's arguments).

### DIFF
--- a/tests/test_install_mender_sh.py
+++ b/tests/test_install_mender_sh.py
@@ -42,6 +42,16 @@ def script_server():
     print("cleaned up script server")
 
 
+@pytest.fixture(scope="function")
+def setup_tester_ssh_connection_f_apt_ready(setup_tester_ssh_connection_f):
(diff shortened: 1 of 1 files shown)


tranchitella (approved):

oleorhagen:
Is it not better to do this in a container, and then retag it as our `latest_raspbian` ?

lluiscampos:
> Is it not better to do this in a container, and then retag it as our `latest_raspbian` ?

That would be an optimization, correct.

I have another change to be submitted soon that makes all these tests run in a generic container so I will get the rid of this fixture. I'll merge as is if the tests pass to unblock your PR, then submit my new rework.

oleorhagen (approved):

PR #121, opened 2021-08-31
test_install_mender_sh.py: Run tests on x86-64 container

lluiscampos:
The purpose of these tests are on verifying the different options of the
install script... and for that we don't need QEMU or ARM emulation.

With this change, we migrate all the tests to use a debian:buster
container, and add *one* test for Raspberry OS.

Locally the tests run now in 10 minutes: 3 for the converted tests and 7
for the new Raspberry OS test.

### DIFF
--- a/tests/test_install_mender_sh.py
+++ b/tests/test_install_mender_sh.py
@@ -17,8 +17,12 @@ import http.server
 import socketserver
 import threading
 import os
-import pytest
+import subprocess
 
+import pytest
+from fabric import Result as FabricResult
+
+SCRIPT_SERVER_ADDR = "localhost"
 SCRIPT_SERVER_PORT = 8000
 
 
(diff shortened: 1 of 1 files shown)


lluiscampos:
@kacf @oleorhagen What do you guys think?

Once I started thinking about writing the JIRA task... I realized it would take the same effort to just do it...

tests/test_install_mender_sh.py
```diff
@@ -43,21 +47,46 @@ def script_server():
...
+def generic_debian_container(request):
+    image = "debian:buster"
+    cmd = "docker run --network host --rm -tid %s" % image
+    output = subprocess.check_output(cmd, shell=True)
+
+    global docker_container_id
+    docker_container_id = output.decode("utf-8").split("\n")[0]
+
+    def finalizer():
+        cmd = "docker stop %s" % docker_container_id
```
kacf, line 59: Should use `docker rm -f`, otherwise stopped containers will accumulate.

tests/test_install_mender_sh.py
```diff
@@ -43,21 +47,46 @@ def script_server():
...
 def check_installed(conn, pkg, installed=True):
     """Check whether the given package is installed on the device given by
     conn."""
 
     res = conn.run(f"dpkg --status {pkg}", warn=True)
-    assert (res.return_code == 0) == installed
+    if isinstance(res, FabricResult):
+        assert (res.return_code == 0) == installed
+    else:
+        assert (res.returncode == 0) == installed
```
kacf, line 89: Clever!

kacf (approved):
Pre approving, but please look below.

oleorhagen (approved):
:ninja: :medal_military: :raised_hands:

tests/test_install_mender_sh.py
lluiscampos: Thanks.

PR #122, opened 2021-09-03
MEN-4759: install-mender.sh: Support installing commercial packages

lluiscampos:
MEN-4759: install-mender.sh: Support installing commercial packages
    
    New flag --commercial to toggle the closed deb packages. It is meant to
    be used only from Hosted Mender and requires --jwt-token argument.
    
    The commercial deb packages are downloaded directly using curl and
    installed using dpkg. Reworked a bit the existing lists of packages in
    the script to have more meaningful purposes.
    
    Other related changes:
    * install-mender.sh: Require bash, which is already required by Mender
      GUI during onboarding.
    * pipeline: Make master/stable copies of packages. This is a short term
      solution but required for the script to have well known URL paths

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -347,6 +347,7 @@ publish:production:s3:scripts:install-mender-sh:
       - production
 
 publish:s3:mender-monitor:
+  stage: publish
   image: debian:buster
   only:
     - master
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -319,8 +319,6 @@ publish:s3:legacy:mender-client:latest:
 
 .publish-template:s3:scripts:install-mender-sh:
   stage: publish
-  dependencies:
-    - test:acceptance
   image: debian:buster
   before_script:
     - apt update && apt install -yyq awscli
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -116,6 +116,18 @@ test:acceptance:
         fi
 
 
+.publish_helper_functions: &publish_helper_functions |
+  # Bash function to check if the string is a final tag
+  function is_final_tag () {
+    version="$1"
+    [[ "${version}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]] && return 0 || return $?
+  }
+  # Bash function to check if the string is a build tag
+  function is_build_tag () {
+    version="$1"
+    [[ "${version}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$-build[0-9]+$ ]] && return 0 || return $?
+  }
+
 publish:s3:apt-repo:
   stage: publish
   only:
@@ -147,16 +159,10 @@ publish:s3:apt-repo:
     - aws s3 cp --recursive s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/dists dists
     - aws s3 cp --recursive s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/pool pool
   script:
-    # Bash function to check if the string is a final tag
-    - |
-      function is_final_tag () {
-        version=$1
-        echo "${version}" | egrep -q '^[0-9]+\.[0-9]+\.[0-9]+$'
-        return $?
-      }
+    - *publish_helper_functions
     # Bash function to compare two semantic versions, returns: 0 (=), 1(>) or 2(<)
     #  * "master" always evaluates to greater than any other version
-    #  * "build" versions like 1.3.4-build for are evaluated as 1.2.3
+    #  * "build" versions like 1.2.3-build4 for are evaluated as 1.2.3
     # Original source: https://stackoverflow.com/a/4025065
     - |
       function vercomp () {
@@ -350,6 +356,17 @@ publish:s3:mender-monitor:
     - build
   before_script:
     - apt update && apt install -yyq awscli
+    - *publish_helper_functions
   script:
     - echo "Publishing mender-monitor version ${MENDER_MONITOR_VERSION} to s3://${S3_BUCKET_NAME_PRIVATE}/${S3_BUCKET_SUBPATH_PRIVATE}/${MENDER_MONITOR_VERSION}/"
     - aws s3 cp output/mender-monitor*.deb s3://${S3_BUCKET_NAME_PRIVATE}/${S3_BUCKET_SUBPATH_PRIVATE}/${MENDER_MONITOR_VERSION}/
+    # Make copies in known destinations to be consumed by get.mender.io script:
+    # * For "experimental" channel, we make a copy of the newest master package in the master directory
+    # * For "stable" channel, we make a copy in a separate "latest" subdirectory of the latest tagged version
+    # This needs to be reworked before Mender 3.2, see MEN-5029 for more details.
+    - if is_build_tag ${MENDER_MONITOR_VERSION} || [ "${MENDER_MONITOR_VERSION}" == "master" ]; then
+    -   aws s3 cp output/mender-monitor*.deb s3://${S3_BUCKET_NAME_PRIVATE}/${S3_BUCKET_SUBPATH_PRIVATE}/master/mender-monitor_master-1_all.deb
+    - fi
+    - if is_final_tag ${MENDER_MONITOR_VERSION}; then
+    -   aws s3 cp output/mender-monitor*.deb s3://${S3_BUCKET_NAME_PRIVATE}/${S3_BUCKET_SUBPATH_PRIVATE}/latest/mender-monitor_latest-1_all.deb
+    - fi
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -1,22 +1,44 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 
 CHANNEL="stable"
 
-AVAILABLE_COMPONENTS="mender-client \
+# Each available component shall be in only one of the lists below
+AVAILABLE_COMPONENTS="\
+mender-client \
 mender-configure \
 mender-configure-demo \
 mender-configure-timezone \
-mender-connect"
+mender-connect \
+mender-monitor \
+"
 
-DEFAULT_COMPONENTS="mender-client \
+DEFAULT_COMPONENTS="\
+mender-client \
 mender-configure \
-mender-connect"
+mender-connect \
+"
 
-DEMO_COMPONENTS="$AVAILABLE_COMPONENTS"
+DEMO_COMPONENTS="\
+mender-configure-demo \
+mender-configure-timezone \
+"
+
+COMMERCIAL_COMPONENTS="\
+mender-monitor \
+"
 
 SELECTED_COMPONENTS="$DEFAULT_COMPONENTS"
+
+# URL prefix from where to download commercial compoments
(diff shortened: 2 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -116,6 +116,20 @@ test:acceptance:
         fi
 
 
+.publish_helper_functions: &publish_helper_functions |
+  # Bash function to check if the string is a final tag
+  function is_final_tag () {
+    version=$1
```
merlin-northern, line 122: suggested change: version=$1 -> version="$1"

.gitlab-ci.yml
```diff
@@ -116,6 +116,20 @@ test:acceptance:
...
+    echo "${version}" | egrep -q '^[0-9]+\.[0-9]+\.[0-9]+$'
+    return $?
+  }
+  # Bash function to check if the string is a build tag
+  function is_build_tag () {
+    version=$1
+    echo "${version}" | egrep -q '^[0-9]+\.[0-9]+\.[0-9]+-build[0-9]+$'
+    return $?
```
merlin-northern, line 130:
```suggestion
[[ "${version}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]
```

.gitlab-ci.yml
```diff
@@ -343,13 +349,25 @@ publish:production:s3:scripts:install-mender-sh:
...
 publish:s3:mender-monitor:
+  stage: publish
   image: debian:buster
   only:
     - master
   dependencies:
     - build
   before_script:
     - apt update && apt install -yyq awscli
+    - *publish_helper_functions
```
merlin-northern, line 360: just for my education, what does `- *` mean?

scripts/install-mender.sh
```diff
@@ -1,23 +1,45 @@
-#!/bin/sh
+#!/bin/bash
```
merlin-northern, line 1: just out of curiosity, have we considered `#!/usr/bin/env bash` here?

scripts/install-mender.sh
```diff
@@ -83,7 +108,25 @@ parse_args() {
                 fi
                 ;;
             --demo)
-                SELECTED_COMPONENTS="$DEMO_COMPONENTS"
+                SELECTED_COMPONENTS="$SELECTED_COMPONENTS $DEMO_COMPONENTS"
```
merlin-northern, line 111: ok, so this virtually means, that commercial components (atm monitor-client) will not be installed with --demo. is that what we want?

scripts/install-mender.sh
```diff
@@ -152,21 +195,82 @@ add_repo() {
...
+       $selected_components_open
+
+    echo "  Success! Please run \`mender setup\` to configure the client."
+}
+
+do_install_commercial() {
+    # Filter commercial components
+    local selected_components_commercial=""
+    for c in $SELECTED_COMPONENTS
+    do
```
merlin-northern, line 228: I do not think we format shell code with do in a separate line.

scripts/install-mender.sh
```diff
@@ -152,21 +195,82 @@ add_repo() {
...
+        if echo "$COMMERCIAL_COMPONENTS" | egrep -q "(^| )$c( |\$)"; then
```
merlin-northern, line 229: this one kills me a bit :)
if we declare the commercial components at the begining as a hash:
```bash
declare -A COMMERCIAL_COMPONENTS_array=([mender-monitor]="1" )
COMMERCIAL_COMPONENTS="${!COMMERCIAL_COMPONENTS_array[@]}"
```
then this can be done with `${COMMERCIAL_COMPONENTS_array[${c}]}`

scripts/install-mender.sh
```diff
@@ -152,21 +195,82 @@ add_repo() {
...
+        version="master"
+    else
+        version="latest"
+    fi
+
+    # Download deb packages
+    for c in $selected_components_commercial
+    do
+        url="$MENDER_COMMERCIAL_DOWNLOAD_URL$(printf ${COMMERCIAL_COMP_TO_URL_PATH_F[$c]} $version $version)"
+        curl -fLsS -H "Authorization: Bearer $JWT_TOKEN" -O $url
```
merlin-northern, line 252: suggested change $url -> "$url"

merlin-northern: shouldnt we check is the `basename $url` exists as a file, i.e.: the donwload was successful, andif not print the relevant error?

scripts/install-mender.sh
```diff
@@ -152,21 +195,82 @@ add_repo() {
...
+    done
+
+    # Install all of them at once and fallback to install missing dependencies
+    local deb_packages_glob=$(echo $selected_components_commercial | sed -e 's/ /*.deb /g; s/$/*.deb/')
+    dpkg --install $deb_packages_glob || apt-get -f install
+
+    # Check individually each package
+    for c in $selected_components_commercial
+    do
```
merlin-northern, line 261: I wonder why we do not use the for loop and do `dpkg -i  "$c"*deb` and then `--status`? do commercial components depend on each other?

scripts/install-mender.sh
```diff
@@ -83,7 +108,25 @@ parse_args() {
...
+                if ! echo "$args_copy" | grep -qe "--jwt-token"; then
+                    echo "ERROR: commercial requires --jwt-token argument."
+                    echo "Aborting."
+                    exit 1
+                fi
+                SELECTED_COMPONENTS="$SELECTED_COMPONENTS $COMMERCIAL_COMPONENTS"
+                ;;
+            --jwt-token)
+                if [ -n "$2" ]; then
+                    JWT_TOKEN=$2
```
merlin-northern, line 123: suggested change: JWT_TOKEN=$2 -> JWT_TOKEN="$2"

scripts/install-mender.sh
```diff
@@ -83,7 +108,25 @@ parse_args() {
                 fi
                 ;;
             --demo)
-                SELECTED_COMPONENTS="$DEMO_COMPONENTS"
+                SELECTED_COMPONENTS="$SELECTED_COMPONENTS $DEMO_COMPONENTS"
+                ;;
+            --commercial)
+                if ! echo "$args_copy" | grep -qe "--jwt-token"; then
```
merlin-northern, line 114:
```suggestion
                if [[ ! "$args" == *"--jwt-token"* ]]; then
```

.gitlab-ci.yml
```diff
@@ -116,6 +116,20 @@ test:acceptance:
         fi
 
 
+.publish_helper_functions: &publish_helper_functions |
+  # Bash function to check if the string is a final tag
+  function is_final_tag () {
+    version=$1
+    echo "${version}" | egrep -q '^[0-9]+\.[0-9]+\.[0-9]+$'
+    return $?
```
kacf, line 124: This does not work if `set -e` is enabled. I would suggest:

```bash
echo "${version}" | egrep -q '^[0-9]+\.[0-9]+\.[0-9]+$' && return 0 || return $?
```

.gitlab-ci.yml
kacf, line 130: Same here ref `set -e`.

kacf (commented):
I did not go very in depth, I think @merlin-northern should have the final review.

scripts/install-mender.sh
lluiscampos: That is a good question. On the conceptual side, you are right in questioning this as "demo" should include all the packages (iow, include commercial packages). But on the implementation side, that would mean that "demo" cannot be used out of Hosted Mender (the JWT would be required always).

That is the reason for which I decided to set it up separately.

scripts/install-mender.sh
lluiscampos: Thanks

scripts/install-mender.sh
lluiscampos: Thanks

lluiscampos: Applied also to `CHANNEL` above

scripts/install-mender.sh
lluiscampos: I know. I based this syntax on the existing code... but once you comment I'll fix the old code too.

scripts/install-mender.sh
lluiscampos: That would work for this check and the one in `do_install_open` (by using opposite logic), but not on the analogous check in `usage` for `DEFAULT_COMPONENTS` (as these are not an array).

For that reason (consistency) and the fact that is pre-existing code, I'll leave as it is.

scripts/install-mender.sh
lluiscampos: :+1:

scripts/install-mender.sh
lluiscampos: > I wonder why we do not use the for loop and do `dpkg -i "$c"*deb` and then `--status`? do commercial components depend on each other? 

That is what I am considering, that commercial components _could_ depend on each other in the future, and then the order on which the user specifies them will make the dpkg call pass or fail

> technically you can drop --status, if -i was successful, right?

Not really. Even if they were independent on each other, some packages might have extra dependencies (for example `mender-monitor` depends on `lmdb-utils`) and hence we need to catch and error and try amending dependencies with `|| apt-get -f install`. And then, the only correct way to check that the pacakge was insalled is with `--status`.

I couldn't find a way to specify up front to dpkg to install the dependencies. Feel free to read its `man` page :stuck_out_tongue:

.gitlab-ci.yml
merlin-northern: yeah, part of the argument, `[[` has nothing to do with `set -e`; `[[` is a shell built-in, as opposed to `[` being a command.

scripts/install-mender.sh
merlin-northern: just an observation, I leave it at your discretion.

scripts/install-mender.sh
merlin-northern: since we are discussing today the onboarding release testing, perhaps we shall also briefly discuss that?

scripts/install-mender.sh
merlin-northern: ok, but this is a killer, perhaps let's create a small task to refactor that?

.gitlab-ci.yml
lluiscampos: Ah, I see what you mean now. We could use `[[` and still do `... ]] && return 0 || return $?`.

scripts/install-mender.sh
lluiscampos: It is related, yes, so we can discuss it. I have an strong opinion though, in that the on-boarding script shall work for OS...

merlin-northern (approved):
pre-approving, thanks for applying many of the comments. the remaining ones are left at your discretion, one we can discuss today with @kacf when we talk about release-testing commercial components.

scripts/install-mender.sh
lluiscampos: Why do you think is that bad? It is just okey for an script that is executed once in a device lifetime, and makes it easier to maintain.

This is the place to discuss and agree, I am not creating a task to refactor something so small and then discuss in the task if we want to do it or not...

merlin-northern: ok, for the sake of continuing, lets keep it as is. it just seemed so suboptimal. so much CPU power to just check a simple thing: if we selected commercial components.

PR #125, opened 2021-09-09
MEN-4759: Support user/pass authentication for commecial components

lluiscampos:
So that an on-prem Mender Enterprise can also download them. For hosted
Mender, both methods are allowed.

Introduces new flags `--auth` and `--username`.


lluiscampos:
Discussed on SU today - we won't implement this.

PR #126, opened 2021-09-09
MEN-4452: Support running mender setup with user specified args

lluiscampos:
This gives the opportunity to the Mender GUI to offload the setup
    details to the install script.
    
    Two main changes:
    * Run mender setup with args passed after `--` marker.
    * Configure mender-connect for RPi on demo mode.

### DIFF
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -57,13 +57,18 @@ Running the Mender installation script.
 }
 
 usage() {
-    echo "usage: install-mender.sh [-h] [-c channel] [component...]"
+    echo "usage: install-mender.sh [options] [component...] [-- [options-for-mender-setup] ]"
+    echo ""
+    echo "options: [-h, help] [-c channel] [--demo] [--commercial]"
     echo "  -h, --help          print this help"
     echo "  -c CHANNEL          channel: stable(default)|experimental"
     echo "  --demo              use defaults appropriate for demo"
     echo "  --commercial        install commercial components, requires --jwt-token"
     echo "  --jwt-token TOKEN   Hosted Mender JWT token"
-    echo "  <component>         list of components to install"
+    echo ""
+    echo "If no components are specified, defaults will be installed"
+    echo ""
+    echo "Anything after a '--' gets passed directly to 'mender setup' command."
     echo ""
     echo "Supported components (x = installed by default):"
     for c in $AVAILABLE_COMPONENTS; do
@@ -88,8 +93,7 @@ is_known_component() {
 parse_args() {
     local selected_components=""
     local args_copy="$@"
-    while [ $# -gt 0 ]
-    do
+    while [ $# -gt 0 ]; do
         case $1 in
             -h|--help)
                 usage
@@ -106,6 +110,7 @@ parse_args() {
                 fi
                 ;;
             --demo)
+                DEMO="1"
                 SELECTED_COMPONENTS="$SELECTED_COMPONENTS $DEMO_COMPONENTS"
                 ;;
             --commercial)
@@ -125,6 +130,11 @@ parse_args() {
                     echo "Aborting."
                     exit 1
                 fi
+                ;;
+            --)
+                shift
+                MENDER_SETUP_ARGS="$@"
+                break
                 ;;
             *)
                 if is_known_component "$1"; then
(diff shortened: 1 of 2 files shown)


lluiscampos:
This is quite open at the moment.

Originally we wanted an explicit new flag "--setup-for-hosted-mender" (see [JIRA](https://tracker.mender.io/browse/MEN-4452)), but when I started working on it I saw that we would need to add several flag to cover the several use cases, in a way duplicating the `mender setup` interface... so concluded that the best is to just accept any argument and pass it downstream. See the [GUI PR](https://github.com/mendersoftware/gui/pull/1978) for how is meant to be used.

scripts/install-mender.sh
```diff
@@ -266,12 +281,48 @@ do_install_commercial() {
...
+}
+
+do_setup_addons() {
+    # Setup for mender-connect, hardcoded for Raspbian OS
+    if [[ "$SELECTED_COMPONENTS" == *"mender-connect"* ]]; then
+        if [ "$DEMO" -eq 1 ]; then
+            echo "  Setting up mender-connect for demo"
+            cat > /etc/mender/mender-connect.conf << EOF
+{
+  "User": "pi",
```
kacf, line 308: I know you said it's hardcoded for RPi, but I'd say that "root" is a better default overall, at least for demo. It will work everywhere. Or is there a specific reason we should use "pi"?

kacf (approved):
I approve, but take a look at my comment. It's worth at least thinking about.

scripts/install-mender.sh
lluiscampos: I have no good reason, maybe `root` is a good idea too. I used `pi` basically because we had it [in the GUI code](https://github.com/mendersoftware/gui/pull/1978/files#diff-68dd097fe57104078af21614d0402539c2377c35e2f44881225afff6a555a751L2011).

Let's see if others here have opinions, otherwise I suggest changing it to `root` but adding a log message here saying something like `Configuring mender-connect for user "root" and shell "bash"...`

scripts/install-mender.sh
```diff
@@ -88,8 +93,14 @@ is_known_component() {
 parse_args() {
     local selected_components=""
     local args_copy="$@"
-    while [ $# -gt 0 ]
-    do
+    local passthrough=0
+    MENDER_SETUP_ARGS=""
+    while [ $# -gt 0 ]; do
+        if test $passthrough -eq 1; then
+            MENDER_SETUP_ARGS="$MENDER_SETUP_ARGS $1"
```
oleorhagen, line 100: Question:
Once `passthrough` is 1, can it ever not be?
If so, then it's not really needed to loop in order to collect the remaining arguments right?

scripts/install-mender.sh
```diff
@@ -266,12 +281,48 @@ do_install_commercial() {
...
+        return
+    fi
+
+    # Return if no setup options were passed
+    if [ -z "$MENDER_SETUP_ARGS" ]; then
+        return
+    fi
+
+    echo "  Setting up mender with options: $MENDER_SETUP_ARGS"
+    mender setup $MENDER_SETUP_ARGS
```
oleorhagen, line 296: Suggestion:
Quote the `SETUP_ARGS` here

scripts/install-mender.sh
```diff
@@ -266,12 +281,48 @@ do_install_commercial() {
...
+    pidof systemd && systemctl restart mender-client
```
oleorhagen, line 297: Q:
What is the pidof for?
Isn't systemd always pid einz? :thinking:

scripts/install-mender.sh
```diff
@@ -266,12 +281,48 @@ do_install_commercial() {
...
+    echo "  Success!"
```
oleorhagen, line 298: Suggestion:
echo "\tSuccess!" instead?

oleorhagen (approved):

scripts/install-mender.sh
```diff
@@ -57,13 +57,18 @@ Running the Mender installation script.
 }
 
 usage() {
-    echo "usage: install-mender.sh [-h] [-c channel] [component...]"
+    echo "usage: install-mender.sh [options] [component...] [-- [options-for-mender-setup] ]"
+    echo ""
+    echo "options: [-h, help] [-c channel] [--demo] [--commercial]"
     echo "  -h, --help          print this help"
     echo "  -c CHANNEL          channel: stable(default)|experimental"
     echo "  --demo              use defaults appropriate for demo"
```
mzedel, line 65: is there a chance to reuse this setting in the setup args? it does look a bit erroneous to have `--demo` twice in [here](https://github.com/mendersoftware/gui/pull/1978/files#diff-0b474816e57448323f2fe953ebd8a228af18faa82b2650929f25d52f6a0c3235R113)
if we would align on setting `--demo` for hosted & on-prem enterprise I think it would be as easy as appending `--demo` to the `$MENDER_SETUP_ARGS`

mzedel (approved):
Should be good as it is and feels so good to see that part removed from the GUI 😅 Thank you!

scripts/install-mender.sh
lluiscampos: Yes, you are right. We can terminate the processing at once after `--` and assign the remaining arguments to `MENDER_SETUP_ARGS` :+1:

scripts/install-mender.sh
lluiscampos: I added this as a simple way to check that `systemd` is running. Specifically, it is not running on tests so we skip calling `systemctl`

scripts/install-mender.sh
lluiscampos: Maybe. I was just following the current styling... I think I'll leave it as is.

scripts/install-mender.sh
lluiscampos: I'll do, thanks.

scripts/install-mender.sh
lluiscampos: Yeah, I know and I thought about it, but unfortunately `--demo` means different things in the script and in `mender setup` - for the Enterprise setup we want "script demo" but "production setup". See [related comment](https://github.com/mendersoftware/gui/pull/1978#discussion_r706039700).

scripts/install-mender.sh
lluiscampos: Done

scripts/install-mender.sh
oleorhagen: Oh, I see, makes sense!
haha, did not strike me at all :see_no_evil: 
Still learning :seedling:

scripts/install-mender.sh
lluiscampos: I had to revert this, when `MENDER_SETUP_ARGS` they are passed a single argument to `mender setup`, resulting in the error: `flag provided but not defined: -demo --device-type cool-device --hosted-mender --tenant-token my-secret-token`

PR #128, opened 2021-09-13
QA-330: Clean up leftovers in the installation process

oleorhagen:


oleorhagen:
Requires:
* https://github.com/mendersoftware/mender-configure-module/pull/29
* https://github.com/mendersoftware/mender/tree/feature-unified_comms (merged to master)

Merged, and cherry picked to all release branches  before merging.

lluiscampos (changes requested):
Thanks for the clean-up :raised_hands: 

I would vote for not back-porting the `Makefile` changes to the old release branches. So in that case we need to:
* Keep the changes in `recipes/mender-client/debian-master`
* Make a copy of the old `recipes/mender-client/debian-master` to `recipes/mender-clien/debian-3.0.x`
* Similarly to `mender-configure`, copying to recipes/mender-configure/debian-1.0.x`

For `mender-connect` should be fine to clean for all versions.

lluiscampos:
> Merged, and cherry picked to all release branches before merging.

To be explicit: the cherry-pick here is what I am voting against.

>     * Make a copy of the old `recipes/mender-client/debian-master` to `recipes/mender-clien/debian-3.0.x`

For the record: if we don't do it before the release, here will be `debian-3.1.x`

PR #129, opened 2021-09-14
QA-322: Align pipeline with Mender release process

lluiscampos:
The idea is two have two sets of publish jobs: one manual and one
    automatic controlled via env variable.
    
    The old SKIP_TESTS variable is superseded by the new one, just with
    inverse logic.

Additionally:
* pipeline: publish: Remove legacy "latest" copy and rename docs publish

    Job publish:s3:legacy:mender-client:latest would make a copy for old
    versions of mender-convert to install the client. Last version to use
    this method was mender-convert 2.2.x, now EOL
    
    On the other hand, the job lish:s3:legacy:mender-client had a misleading
    name, as the published package is still in use for direct download from
    the Mender Docs site, so rename it to be explicit.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -14,14 +14,16 @@ variables:
   S3_BUCKET_REPO_PATH: "repos/debian"
   # Scripts folder subpath
   S3_BUCKET_SCRIPTS_PATH: "repos/scripts"
-  # Set to true to skip tests (i.e. old Mender client build). See QA-123
-  SKIP_TESTS: "false"
   # Update the 'latest' debian alias to a new client release
   PUBLISH_LATEST_CLIENT_PACKAGE: "false"
   # GPG keys for build and distribution, to be set by CI/CD variables
   GPG_PRIV_KEY_BUILD: ""
   GPG_PUB_KEY_BUILD: ""
   GPG_PRIV_KEY_DIST: ""
+  # Whether to run acceptance tests.
+  TEST_MENDER_DIST_PACKAGES: "true"
+  # Whether to publish packages automatically - they can always be published manually
+  PUBLISH_MENDER_DIST_PACKAGES_AUTOMATIC: "false"
 
 include:
   - project: 'Northern.tech/Mender/mendertesting'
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -355,3 +368,12 @@ publish:s3:mender-monitor:
...
     -   aws s3 cp output/mender-monitor*.deb s3://${S3_BUCKET_NAME_PRIVATE}/${S3_BUCKET_SUBPATH_PRIVATE}/latest/mender-monitor_latest-1_all.deb
     - fi
+
+publish:s3:mender-monitor:manual:
+  when: manual
+  extends: .template:publish:s3:mender-monitor
+
+publish:s3:mender-monitor:automatic:
+  rules:
+    - if: '$PUBLISH_MENDER_DIST_PACKAGES_AUTOMATIC == "true"'
```
oleorhagen, line 378: Just a quick question:
Reading this: https://docs.gitlab.com/ee/ci/jobs/job_control.html#specify-when-jobs-run-with-rules
I now expect this can be done in one job, like:
```
rules:
  - if: "....." == "true"
  - if: "..." != "true"
    when: manual
```

oleorhagen (approved):
Looks good, just left one question we can discuss

.gitlab-ci.yml
lluiscampos: That is true, but I think is much better to have the explicit `:manual` `:automatic` name in the job.

I got used to it in `mender-qa` (since you introduced it, I think) and is an easy way to quickly check what is going to be published the moment you trigger the pipeline, so that you don't end up waiting for it and then unexpectedly gets published.

Unless you insist I'll leave it as it is.

kacf (approved):

lluiscampos:
New commit:

    pipeline: apt-repo: Do not override existing packages

    It could happen that a given release re-uses an old component already
    tagged and published and we need to skip copying it. Otherwise we are
    corrupting the APT repository indexes.

    In any case shall we replace a package, the manual copy is meant only to
    workaround the publishing of bugfix releases, where reprepro refuses to
publish it for us due to the versioning scheme.

lluiscampos:
New commit:

    pipeline: apt-repo: Check newer versions on correct codename
    
    So that a build tag (which goes to experimental) can potentially
    re-built without breaking the pipeline.

kacf (approved):

lluiscampos:
Reworked to skip all tests conditionally. [Here](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/372631509) the normal build, [here](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/372632336) the one disabling tests.

PR #131, opened 2021-09-20
MEN-5075: install-mender.sh: Fix script when not using --demo

lluiscampos:
If DEMO is not defined, the script would fail with:
`line 298: [: : integer expression expected`


oleorhagen (approved):

oleorhagen:
Alternative would be to use the string `true` false semantics I guess?

lluiscampos:
> Alternative would be to use the string `true` false semantics I guess?

Yes, that would evaluate correctly even on empty string.

PR #132, opened 2021-09-20
pipeline: apt-repo: Correctly evaluate build tags and git versions

lluiscampos:
Amends commit 9741aac.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -296,12 +296,8 @@ test:acceptance:
 
     # Upload to bucket
     - aws s3 sync db s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/db
-    - aws s3 sync pool s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/pool
-    - aws s3 ls --recursive s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/pool/ |
-      awk '{cmd="aws s3api put-object-acl --acl public-read --bucket ${S3_BUCKET_NAME} --key "$4; system(cmd)}'
-    - aws s3 sync dists s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/dists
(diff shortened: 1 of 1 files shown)


lluiscampos:
Follow-up from https://github.com/mendersoftware/mender-dist-packages/pull/129#issuecomment-920802169

oleorhagen (approved):

.gitlab-ci.yml
```diff
@@ -217,12 +217,22 @@ test:acceptance:
...
+        if [[ "${current_version}" == *"~git"* ]]; then
+          # Package is master, assume $package_version is "newer"
+          return 0
+        fi
+
+        # Remove debian -1 suffix
+        current_version_semver=$(echo $current_version | sed 's/-1$//' )
+        # Handle build versions for comparison
+        current_version_comp=$(echo $current_version_semver | sed 's/-build/\./' )
+        package_version_comp=$(echo $package_version | sed 's/-build/\./' )
```
kacf, line 233: I think it was like this already, so I'm probably wrong, but I'm asking for my own curiosity: Won't this always compare "-build123" builds higher than finals? "1.2.3-build4" will be translated to "1.2.3.4", which is higher than "1.2.3", right?

lluiscampos: That is correct.

However, we will never run into that case, as we are comparing either stable versions (always final tags) or experimental versions (combination of build tags or git tags).

For the latter, there will be three use cases:
* First build tag: 1.2.3-build1 vs 1.2.3~gitSOMETHING -> especial early handling, returns 1.2.3-build1 being newer
* Other build tags: 1.2.3-build2 vs 1.2.3-build1 -> translation as you stated, evaluates as 1.2.3-build2 being newer
* Fist git tag: master vs 1.2.3-build2 -> especial handling in `vercomp`, evaluates as master being newer

(there is quite a bit of detail here, sorry. I made local experiments and put good time in thinking this thorough, and I hope I got it right :crossed_fingers: )

PR #134, opened 2021-09-25
pipeline: Publish and reuse Docker images

lluiscampos:
To always build the Docker images from scratch is just a waste... This
commit reworks the pipeline so that the images are only built on changes
to the relevant files; and the packages are build either pulling the
images from upstream or loading the local images.

The docker-build-package script has been reworked to take the package to
build as an input, with the intention in the future to select via
pipeline variables which packages to build.


lluiscampos:
This PR is born out of frustration... I had to retry [the Docker images build](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/jobs/1622653674) five times (so far) during the Mender 3.1.0-build6 due to errors in upstream sources.

What do you guys think? Does it make sense?

lluiscampos:
(I have syntax errors in the GitLab pipeline file, but I won't fix it until I get some initial feedback on the approach).

oleorhagen (approved):

kacf (approved):
I think it makes sense. I couldn't spot anything wrong, but Gitlab will be the ultimate judge.

lluiscampos:
Thanks for the positive feedback I force-pushed the pipeline fixes (and rebased on top of latest mater)

lluiscampos:
The new commit makes use of https://github.com/mendersoftware/mendertesting/pull/158 to save/load the huge images. I'll merge after green build :crossed_fingers:

PR #142, opened 2021-10-06
mender-deb-package: Modify logic to detect private repo URL

lluiscampos:
After the rework from 3eea53c, MENDER_PRIVATE_REPO_ACCESS_USER is
defined for all calls to mender-deb-package, so this if block needs to
be a bit smarter to only activate when it is actually in use (iow when
the user string is in the repository URL).

### DIFF
--- a/mender-deb-package
+++ b/mender-deb-package
@@ -37,7 +37,7 @@ verify_script_arguments() {
 checkout_repo_clean_local_path() {
   local -r repo_url="$1"
   local repo_path="${repo_url#https://}"
-  if [[ -n "${MENDER_PRIVATE_REPO_ACCESS_USER}" ]]; then
+  if [[ "$repo_url" == *"${MENDER_PRIVATE_REPO_ACCESS_USER:-none}"* ]]; then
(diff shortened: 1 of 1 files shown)


mender-deb-package
```diff
@@ -37,7 +37,7 @@ verify_script_arguments() {
 checkout_repo_clean_local_path() {
   local -r repo_url="$1"
   local repo_path="${repo_url#https://}"
-  if [[ -n "${MENDER_PRIVATE_REPO_ACCESS_USER}" ]]; then
+  if [[ "$repo_url" == *"${MENDER_PRIVATE_REPO_ACCESS_USER:-none}"* ]]; then
```
oleorhagen, line 40: Question:
If the `MENDER_PRIVATE_REPO_ACCESS_USER` is defined for all calls, then why the default `:-none`?

lluiscampos: Good question. It is defined for all calls when running in master, not when running on a PR branch nor locally. (I didn't had the `none` default in my first attempt and then ironically the build fail on the PR build).

PR #143, opened 2021-10-06
pipeline: publish:images: Add missing docker login

lluiscampos:


.gitlab-ci.yml
```diff
@@ -219,6 +219,7 @@ publish:images:
     - docker load -i builder-arm64.tar
     - docker load -i builder-armhf.tar
     - docker load -i builder-amd64.tar
+    - docker login -u $DOCKER_HUB_USERNAME -p $DOCKER_HUB_PASSWORD
```
merlin-northern, line 222: I always wonder, do we need double quotes around variables here?

lluiscampos: Copied from [the template](https://github.com/mendersoftware/mendertesting/blob/master/.gitlab-ci-check-docker-build.yml#L47). As long as the password have no spaces, and that it cannot expand to a file in the current dir... it should be okey.

lluiscampos: Launch a PR in the template, then :wink:

merlin-northern: https://github.com/mendersoftware/mendertesting/pull/161 you know how to make me file a PR

PR #144, opened 2021-10-25
install-mender.sh: Propagate --demo flag into mender setup

lluiscampos:
Translated accordingly. See MEN-5138.

### DIFF
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -282,13 +282,18 @@ do_setup_mender() {
         return
     fi
 
-    # Return if no setup options were passed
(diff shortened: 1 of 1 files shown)


alfrunes:
Psst! @lluiscampos what's the status here?

lluiscampos:
@alfrunes Waiting for the release :)

Basically to not make the release testing too complex (this kind of modification has a chicken-egg problem between the mender client and the install script) I decided to make it in two steps: first release the client in upcoming Mender 3.2, then get this change (and others I have planned for GUI) on Mender 3.3.

PR #149, opened 2021-11-18
mender-deb-package: Amend version increment for "master" packages

lluiscampos:
With only increasing minor, we risk that for a component with last
version `1.2.3` the next master is `1.3.3` which evaluates higher that
the next final release `1.3.0`...

Amends commit 99885d9.


kacf (approved):
Doh! One of us should have seen that! :-)

PR #152, opened 2021-11-24
MEN-5137: Add the example monitors as a part of the standard install

oleorhagen:
Changelog: None


lluiscampos (commented):
Is this one ready to be reviewed?

Missing:
* As we are introducing a breaking change, we need to split now  `mender-monitor` recipes `debian-master` with `debian-1.0.x`, same as it is done for other packages
* the install script needs to be modified to install the new package when `--demo` flag is provided, same as we do for `mender-configure` package(s).

oleorhagen:
@lluiscampos yes, you can have a look. At this point, I'm guessing mostly the names will change in the `monitor-client` repo.

So go ahead.

lluiscampos:
> What's the breaking change we're introducing?

`monitor-client` 1.0.x does not have `install-example-monitors` target in it's `Makefile`. That is a breaking change in the sense that today's `debian-master` recipe won't build tomorrow `1.0.1` release. Hence we need to copy the old contents of `debian-master` into `debian-1.0.x`.

recipes/mender-monitor/debian-master/control
```diff
@@ -13,3 +13,7 @@ Description: Mender Monitor client
  device events to the Mender server, such as log, service files and D-Bus
  signals.
  
+Package: mender-monitor-demo
+Architecture: all
+Depends: mender-monitor, ${misc:Depends}
+Description: Adds demo functionality for the Monitor Client
```
lluiscampos, line 19: Here (and above, in the original package) I would use "Mender Monitor", which is the name of the package and the name we are using in the docs.

recipes/mender-monitor/debian-master/mender-monitor.install
```diff
@@ -0,0 +1,27 @@
...
+usr/share/mender-monitor/lib/ctl-lib.sh
+usr/share/mender-monitor/lib/buf-lib.sh
+usr/share/mender-monitor/lib/fixlenstore-lib.sh
+usr/share/mender-monitor/lib/confpush-lib.sh
+usr/share/mender-monitor/lib/mender-client-lib.sh
+usr/share/mender-monitor/lib/log-lib.sh
+usr/share/mender-monitor/lib/alert-lib.sh
+usr/share/mender-monitor/lib/expiry-lib.sh
+usr/share/mender-monitor/lib/service-lib.sh
+usr/share/mender-monitor/config/config.sh
```
lluiscampos, line 27: Can we do all these a `.dirs` file? I am concern we will later add files in the repo and miss them here.

_EDIT_: This might be okey, see my comment below

recipes/mender-monitor/debian-master/mender-monitor.install
```diff
@@ -0,0 +1,27 @@
+/usr/bin/mender-monitord
+/usr/bin/mender-monitorctl
+
+/etc/mender-monitor/monitor.d/service.sh
+/etc/mender-monitor/monitor.d/log.sh
+/etc/mender-monitor/monitor.d/dbus.sh
```
lluiscampos, line 6: Where are the new demo subsystems?

recipes/mender-monitor/debian-master/rules
```diff
@@ -5,5 +5,8 @@
 %:
 	dh $@
 
-override_dh_link:
-	true
+override_dh_auto_install:
+	DESTDIR=debian/tmp make install install-example-monitors
+
+override_dh_missing:
+	dh_missing --fail-missing
```
lluiscampos, line 12: :thinking: This will catch missing files as I worried about above?

I didn't know about this but sounds like a nice idea! The only drawback is that we will often need to update the recipes and create more `debian-X.X.X` copies.

lluiscampos (changes requested):
Looking good, I see you have been studying some Debian policy :nerd_face: 

A part from what I said earlier (install script + copy to `debian-1.0.x`), bunch of comments below.

recipes/mender-monitor/debian-master/rules
oleorhagen: Yes, I would vote for giving it a try though, I actually was missing a file, after the original install.

What you think?

lluiscampos: Yes, let's do it this way. I like it too.

recipes/mender-monitor/debian-master/control
oleorhagen: That description is already present though, keep it out of this PR?

lluiscampos: You mean the description for `mender-monitor` package? for 1.0.x we should leave it, but for master we should fix it now that we have seen it.

recipes/mender-monitor/debian-master/mender-monitor.install
oleorhagen: Only part of the demo package install

lluiscampos: Exactly, but I don't see them there in `mender-monitor-demo.install`

oleorhagen: Have another look :smile_cat:

lluiscampos: You added them now in the last force-push. Good :+1:

lluiscampos (approved):

PR #158, opened 2021-12-22
test: Add a check for the output from the newer client

oleorhagen:
The new client has changed the route taken for authorization in the state
machine, hence a new output match is needed.

Changelog: None


alfrunes (approved):

lluiscampos:
:+1:

PR #167, opened 2022-01-20
MEN-5400: install-mender.sh: Fix install of demo commercial packages

lluiscampos:
Commercial demo packages need to be filtered out when installing from
repository and filtered in when installing from direct downloads.

### DIFF
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -221,7 +221,7 @@ do_install_open() {
     # Filter out commercial components
     local selected_components_open=""
     for c in $SELECTED_COMPONENTS; do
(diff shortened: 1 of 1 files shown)


oleorhagen (approved):
:+1:

merlin-northern:
if I need --commercial why am I not getting this from the bash snippet in the UI while running demo in enterprise-testing mode?

lluiscampos:
> if I need --commercial why am I not getting this from the bash snippet in the UI while running demo in enterprise-testing mode?

Installing commercial packages from the script is only supported on Hosted Mender.

PR #168, opened 2022-01-21
Add libffi6 as a depency for the client

oleorhagen:
Changelog: None


oleorhagen:
libffi6 is not present on newest `bullseye`, which has `libffi7` installed by default only.

tranchitella (approved):

PR #174, opened 2022-01-25
pipeline: fixes for bugfix release

lluiscampos:
pipeline: publish mender-monitor: Add guards around demo package
    
    Which does not exist for 1.0.x builds

    pipeline: publish commercial: comment out the "latest" updates
    
    The logic is not good enough and would have overridden recently released
    mender-monitor 1.1.0 with soon to be released mender-monitor 1.0.1 (both
    are final tags!)
    
    We are following up in MEN-5029.


oleorhagen (approved):
:+1:

PR #175, opened 2022-01-25
install the missing enabled directory

oleorhagen:
The mender-monitor package does not install the enabled directory.

This is done on mender-monitorctl create. But when running the enabled command,
the directory is not created.

Changelog: None

### DIFF
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -319,6 +319,13 @@ EOF
     fi
 }
 
+do_install_missing_monitor_dirs () {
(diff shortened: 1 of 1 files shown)


lluiscampos (changes requested):
This needs to be done only when installing `mender-monitor-demo`. So please check for that package with something like:

```
if [[ "$SELECTED_COMPONENTS" == *"mender-monitor-demo"* ]]; then
```

oleorhagen:
Follow-up ticket: https://tracker.mender.io/browse/MEN-5409

lluiscampos (approved):

PR #176, opened 2022-01-25
cherry: install the missing enabled directory to production

oleorhagen:


mzedel (approved):

tranchitella (approved):
As mentioned, we need to fix this in the proper way.

PR #177, opened 2022-01-26
feat: update `install-mender.sh` to detect LSB distro/version

tranchitella:
The install-mender.sh is now capable of detecting the LSB distro and
version, and selecting the distro/version-specific APT repository.

Changelog: title

### DIFF
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -202,19 +202,30 @@ get_deps() {
 add_repo() {
     curl -fsSL $REPO_URL/gpg | apt-key add -
 
-    repo="deb [arch=$ARCH] $REPO_URL $CHANNEL main"
-
-    echo "Checking if mender sources already exist in '/etc/apt/sources.list'..."
-    if grep -F "$repo" /etc/apt/sources.list; then
-        echo "Removing the old mender debian source list from /etc/apt/sources.list..."
-        if ! sed -i.bak -e "\,$REPO_URL,d" /etc/apt/sources.list; then
-            echo "Failed to remove the existing mender debian source from '/etc/apt/sources.list'."
-            echo "This probably means that there already exists a source in your sources.list."
-            echo "Please remove it manually before proceeding."
-            exit 1
-        fi
+    local repo_deprecated="deb [arch=$ARCH] $REPO_URL $CHANNEL main"
+    if grep -F "$repo_deprecated" /etc/apt/sources.list >/dev/null; then
+        echo "ERROR: deprecated repository found in apt sources lists."
+        echo "Please remove it manually with: sudo sed -i.bak -e \"\,$repo_deprecated,d\" /etc/apt/sources.list"
+        echo "See https://docs.mender.io for updated APT repos information"
+        exit 1
+    fi
+    if test -f /etc/apt/sources.list.d/mender.list && \
+            grep -F "$repo_deprecated" /etc/apt/sources.list >/dev/null; then
+        echo "ERROR: deprecated repository found in apt sources lists."
+        echo "Please remove it manually with: sudo rm /etc/apt/sources.list.d/mender.list"
+        echo "See https://docs.mender.io for updated APT repos information"
+        exit 1
     fi
 
+    local repo_dist=""
+    if [[ "$LSB_DIST" == "raspbian" ]]; then
+        repo_dist="debian"
+    else
+        repo_dist="$LSB_DIST"
+    fi
+
+    local repo="deb [arch=$ARCH] $REPO_URL $repo_dist/$DIST_VERSION/$CHANNEL main"
+    echo "Installing Mender APT repository at /etc/apt/sources.list.d/mender.list..."
     echo "$repo" > /etc/apt/sources.list.d/mender.list
 }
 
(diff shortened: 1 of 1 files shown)


scripts/install-mender.sh
```diff
@@ -326,7 +326,109 @@ do_install_missing_monitor_dirs () {
...
+    command -v "$@" > /dev/null 2>&1
+}
+
+# Set the LSB_DIST and DIST_VERSION variables guessing the distribution and version;
+# It also checks if this is a forked Linux distro.
+# Credits: https://get.docker.com/
+check_dist_and_version() {
+    # Every system that we officially support has /etc/os-release
+    if [ -r /etc/os-release ]; then
+        LSB_DIST="$(. /etc/os-release && echo "$ID" | tr '[:upper:]' '[:lower:]')"
```
merlin-northern, line 339: suggested change:
```bash
. /etc/os-release && echo "${ID^^}"
```

scripts/install-mender.sh
```diff
@@ -326,7 +326,109 @@ do_install_missing_monitor_dirs () {
...
+    fi
+    case "$LSB_DIST" in
+        ubuntu)
+            if command_exists lsb_release; then
+                DIST_VERSION="$(lsb_release --codename | cut -f2)"
```
merlin-northern, line 344: cut has been known to occasionally change the default -d, could we have explicit -d here?

merlin-northern: actually the safest is: 
```bash
| awk '{print($NF);}'
```

scripts/install-mender.sh
```diff
@@ -326,7 +326,109 @@ do_install_missing_monitor_dirs () {
...
+                ;;
+                *)
+                    echo "ERROR: your distribution's version ($DIST_VERSION) is either not recognized or not supported."
+                    echo "Aborting."
+                    exit 1
+                ;;
+            esac
+        ;;
+        debian|raspbian)
+            DIST_VERSION="$(sed 's/\/.*//' /etc/debian_version | sed 's/\..*//')"
```
merlin-northern, line 364: isnt `lsb_release -r` better?

scripts/install-mender.sh
```diff
@@ -326,7 +326,109 @@ do_install_missing_monitor_dirs () {
...
+        # Check if the `-u` option is supported
+        set +e
+        lsb_release -a -u > /dev/null 2>&1
+        lsb_release_exit_code=$?
+        set -e
+
+        # Check if the command has exited successfully, it means we're in a forked distro
+        if [ "$lsb_release_exit_code" = "0" ]; then
+            # Get the upstream release info
+            LSB_DIST=$(lsb_release -a -u 2>&1 | tr '[:upper:]' '[:lower:]' | grep -E 'id' | cut -d ':' -f 2 | tr -d '[:space:]')
```
merlin-northern, line 397: I am getting:
```
lsb_release: error: no such option: -u
```

tranchitella: As the comment says, it is available on forked distros only.
That's why we try the command, and then check the exit code.

scripts/install-mender.sh
```diff
@@ -326,7 +326,109 @@ do_install_missing_monitor_dirs () {
...
+            DIST_VERSION=$(lsb_release -a -u 2>&1 | tr '[:upper:]' '[:lower:]' | grep -E 'codename' | cut -d ':' -f 2 | tr -d '[:space:]')
+        else
+            if [ -r /etc/debian_version ] && [ "$LSB_DIST" != "ubuntu" ] && [ "$LSB_DIST" != "raspbian" ]; then
+                if [ "$LSB_DIST" = "osmc" ]; then
+                    # OSMC runs Raspbian
+                    LSB_DIST=raspbian
+                else
+                    # We're Debian and don't even know it!
```
oleorhagen, line 405: :smile:

scripts/install-mender.sh
```diff
@@ -326,7 +326,109 @@ do_install_missing_monitor_dirs () {
...
+        ubuntu)
+            if command_exists lsb_release; then
+                DIST_VERSION="$(lsb_release --codename | cut -f2)"
+            fi
+            if [ -z "$DIST_VERSION" ] && [ -r /etc/lsb-release ]; then
+                DIST_VERSION="$(. /etc/lsb-release && echo "$DISTRIB_CODENAME")"
+            fi
+            case "$DIST_VERSION" in
+                focal)
+                    DIST_VERSION="focal"
```
oleorhagen, line 351: Nit:
No-op

oleorhagen (approved):
Tested on ubuntu 20 and bullseye, seems to work

scripts/install-mender.sh
tranchitella: I know, this is mainly for consistency with the Debian's case and to make the code clear.

tranchitella:
@merlin-northern the code you reviewed is from the Docker script. If you don't mind, I would avoid changing it to keep it consistent with the "upstream" version.

scripts/install-mender.sh
```diff
@@ -202,7 +202,7 @@ get_deps() {
 add_repo() {
     curl -fsSL $REPO_URL/gpg | apt-key add -
 
-    repo="deb [arch=$ARCH] $REPO_URL $CHANNEL main"
+    repo="deb [arch=$ARCH] $REPO_URL $DIST_VERSION/$CHANNEL main"
```
lluiscampos, line 205: As we informally agreed on Slack, we need here first to check if the old repo exists in `/etc/apt/sources.list.d/mender.list`. If so, we print an error message and give instructions to how to manually remove it (`sudo rm /etc/apt/sources.list.d/mender.list`) and ask to run again the script.

lluiscampos (changes requested):
Everything else looks good!

scripts/install-mender.sh
lluiscampos, line 205: Also, this is wrong. The codenames for our distros are:
* `debian/buster/experimental`
* `debian/buster/stable`
* `debian/bullseye/experimental`
* `debian/bullseye/stable`
* `ubuntu/bionic/experimental`
* `ubuntu/bionic/stable`
* `ubuntu/focal/experimental`
* `ubuntu/focal/stable`

So here we need to do something like:
```suggestion
    repo="deb [arch=$ARCH] $REPO_URL $LSB_DIST/$DIST_VERSION/$CHANNEL main"
```

And below we need to override `LSB_DIST` to `debian` when `raspbian` is detected.

scripts/install-mender.sh
```diff
@@ -202,19 +202,29 @@ get_deps() {
 add_repo() {
```
lluiscampos, line 202: This function is what I have modified in the last force-push. @merlin-northern @tranchitella could you please review it?

lluiscampos (approved):

scripts/install-mender.sh
```diff
@@ -202,19 +202,29 @@ get_deps() {
...
-    if grep -F "$repo" /etc/apt/sources.list; then
-        echo "Removing the old mender debian source list from /etc/apt/sources.list..."
-        if ! sed -i.bak -e "\,$REPO_URL,d" /etc/apt/sources.list; then
-            echo "Failed to remove the existing mender debian source from '/etc/apt/sources.list'."
-            echo "This probably means that there already exists a source in your sources.list."
-            echo "Please remove it manually before proceeding."
-            exit 1
-        fi
+    repo_deprecated="deb [arch=$ARCH] $REPO_URL $CHANNEL main"
+    if grep -F "$repo_deprecated" /etc/apt/sources.list; then
```
tranchitella, line 206: Maybe redirect stdout to /dev/null to avoid polluting the console.

```suggestion
    if grep -F "$repo_deprecated" /etc/apt/sources.list >/dev/null; then
```

Or, if you prefer, use `-q`:

```suggestion
    if grep -q -F "$repo_deprecated" /etc/apt/sources.list; then
```

scripts/install-mender.sh
```diff
@@ -202,19 +202,29 @@ get_deps() {
...
-    echo "Checking if mender sources already exist in '/etc/apt/sources.list'..."
-    if grep -F "$repo" /etc/apt/sources.list; then
-        echo "Removing the old mender debian source list from /etc/apt/sources.list..."
-        if ! sed -i.bak -e "\,$REPO_URL,d" /etc/apt/sources.list; then
-            echo "Failed to remove the existing mender debian source from '/etc/apt/sources.list'."
-            echo "This probably means that there already exists a source in your sources.list."
-            echo "Please remove it manually before proceeding."
-            exit 1
-        fi
+    repo_deprecated="deb [arch=$ARCH] $REPO_URL $CHANNEL main"
```
merlin-northern, line 205: should be local

scripts/install-mender.sh
```diff
@@ -202,19 +202,29 @@ get_deps() {
...
+        echo "See https://docs.mender.io for updated APT repos information"
+        exit 1
     fi
 
+    if [[ "$LSB_DIST" == "raspbian" ]]; then
+        repo_dist="debian"
+    else
+        repo_dist="$LSB_DIST"
+    fi
+    repo="deb [arch=$ARCH] $REPO_URL $repo_dist/$DIST_VERSION/$CHANNEL main"
```
merlin-northern, line 225: should be local

scripts/install-mender.sh
```diff
@@ -202,19 +202,29 @@ get_deps() {
...
-            echo "This probably means that there already exists a source in your sources.list."
-            echo "Please remove it manually before proceeding."
-            exit 1
-        fi
+    repo_deprecated="deb [arch=$ARCH] $REPO_URL $CHANNEL main"
+    if grep -F "$repo_deprecated" /etc/apt/sources.list; then
+        echo "ERROR: deprecated repository found in apt sources lists."
+        echo "Please remove it manually with: sudo sed -i.bak -e \"\,$repo_deprecated,d\" /etc/apt/sources.list"
+        echo "See https://docs.mender.io for updated APT repos information"
+        exit 1
```
merlin-northern, line 210: are we sure about exits here? or should be returns?

scripts/install-mender.sh
```diff
@@ -202,19 +202,29 @@ get_deps() {
...
+    if test -f /etc/apt/sources.list.d/mender.list && \
+            grep -F "$repo_deprecated" /etc/apt/sources.list; then
+        echo "ERROR: deprecated repository found in apt sources lists."
+        echo "Please remove it manually with: sudo rm /etc/apt/sources.list.d/mender.list"
+        echo "See https://docs.mender.io for updated APT repos information"
+        exit 1
     fi
 
+    if [[ "$LSB_DIST" == "raspbian" ]]; then
+        repo_dist="debian"
```
merlin-northern, line 221: should be local

scripts/install-mender.sh
merlin-northern: could we have double quotes?

scripts/install-mender.sh
lluiscampos: Thanks

scripts/install-mender.sh
lluiscampos: We should exit because if we don't install the repo we just cannot install the packages...

lluiscampos:
The tests fail due to wrong smoke build of the packages. Fix here:
* https://github.com/mendersoftware/mender-dist-packages/pull/180

PR #179, opened 2022-01-27
MEN-5410: Build packages for Debian and Ubuntu two latest releases

lluiscampos:
MEN-5410: Build packages for Debian and Ubuntu two latest releases
    
    Our policy of building from "oldstable" and support two releases this
    way has got now into trouble with the upgrade from libffi6 to libffi7 in
    latest Debian bullseye.
    
    This commit reworks many parts of the build process to build and publish
    deb packages for the last two releases of Debian (Buster and Bullseye)
    and the last two of Ubuntu (Bionic and Focal).
    
    Notable changes:
    * Dockerfile: adapt to work on the four base imgages. Note that Ubuntu
      does not support foreign architectures in their main repository and a
      ports.ubuntu.com needs to be added. Note also the special handling for
      libffi6/libffi7
    * docker-build- wrapper scripts: Modify to build and use the 2 distros x
      2 releases x 3 architectures = 12 builder images
    * .gitlab-ci.yml: Adapt to the above accordingly
    * recipes and mender-deb-package script: Move away from the hardcoded
      Depends (which brought the breaking problem in the first place) and
      use instead Debian tools to figure out the dependencies. For the case
      where we use our custom toolchain to support ARMv6 (this is, Debian
      for arm 32 bits) post-process the recipes to get the hardcoded
      pre-generated dependencies.
    
    Some minor clean-ups in the recipes were also done.

.

    MEN-5410: Refactor scritps and pipeline to build in parallel
    
    This greatly reduces the time and disk usage of the CI slaves

.

    MEN-5410: Append distribution and release to the package version
    
    The original approach had the limitation that the produced deb packages
    will have the same version (and name) on different platforms, which will
    error to update the repository (all distributions in our repo share the
    same pool).
    
    Removing the tentative directory hierarchy also helped with the unknown
    on publishing, the acceptance tests and reduces risk on the changes to
    be made for the reprepro wrapper script.


lluiscampos:
Although I resolved it in three steps (hence three commits) I think it is best reviewed in one go - the last commit reverts a part of the first commit.

mender-deb-package
```diff
@@ -50,13 +50,19 @@ checkout_repo() {
...
   # Create a version from Git.
-  #  - For Git tags: X.Y.Z-1
-  #  - For master: X.Y.Z~git<commit-date>.<commit-sha>-1+b<BUILD_ID>
+  #  - For Git tags: X.Y.Z-1+DISTRO+CODENAME
+  #  - For master: X.Y.Z~git<commit-date>.<commit-sha>-1+DISTRO+CODENAME+b<BUILD_ID>
   #     where X.Y.Z is latest tag (not necessarily matching git describe)
+  debian_suffix="1+$OS_DISTRO+$OS_CODENAME"
   if [ "$VERSION" != "master" ] && git describe --tags --exact-match 2>/dev/null; then
-    DEB_VERSION="$(git describe --tags --exact-match)-1"
+    DEB_VERSION="$(git describe --tags --exact-match)-$debian_suffix"
```
lluiscampos, line 65: Note to self, `prepare_recipe` below needs fixing following this change.

Dockerfile
```diff
@@ -1,18 +1,31 @@
...
-RUN apt-get update && \
+RUN export DEBIAN_FRONTEND=noninteractive && \
+    apt-get update && \
     apt-get install -y \
     build-essential \
     git wget curl \
     debhelper devscripts
 
 ARG ARCH=amd64
+ARG DISTRO=debian
```
oleorhagen, line 13: Double ARG definition, right?

https://github.com/mendersoftware/mender-dist-packages/pull/179/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R1

docker-build-image
```diff
@@ -0,0 +1,46 @@
...
+ARCH="${3}"
+
+# Pass GPG key as build argument if set
+if [ -n "${GPG_KEY_BUILD}" ]; then
+    docker build \
+            --tag ${IMAGE_NAME_PREFIX}-${DISTRO}-${RELEASE}-${ARCH} \
+            --build-arg DISTRO=${DISTRO} \
+            --build-arg VERSION=${RELEASE} \
+            --build-arg ARCH=${ARCH} \
+            --build-arg GPG_KEY_BUILD="${GPG_KEY_BUILD}" \
```
oleorhagen, line 37: Do as you wish with suggestion:
Can do `{GPG_KEY_BUILD:+--build-arg GPG_KEY_BUILD="${GPG_KEY_BUILD}"}`

mender-deb-package
lluiscampos: Fixed

mender-deb-package
```diff
@@ -50,13 +50,27 @@ checkout_repo() {
...
+  else
+    DEB_DISTRIBUTION="experimental"
+  fi
+}
+
 get_deb_version() {
   # Create a version from Git.
-  #  - For Git tags: X.Y.Z-1
-  #  - For master: X.Y.Z~git<commit-date>.<commit-sha>-1+b<BUILD_ID>
+  #  - For Git tags: X.Y.Z-1+DISTRO+CODENAME
```
oleorhagen, line 68: nit:
`X.Y.Z-<debian_suffix>...`

oleorhagen (approved):

oleorhagen:
:100:

recipes/mender-client/debian-master/control
```diff
@@ -8,13 +8,15 @@ Build-Depends: debhelper (>= 9)
 Package: mender-client
 Conflicts: mender
 Architecture: any
-Depends: libc6 (>= 2.12), libglib2.0-0 (>=2.50), liblzma5 (>= 5.1), libssl1.1 (>= 1.1), libffi6 (>= 3.2)
+Depends: ${shlibs:Depends}, ${misc:Depends}
+# debian bullseye Depends: libc6 (>= 2.14), libglib2.0-0 (>= 2.26.0), liblzma5 (>= 5.1.1alpha+20120614), libssl1.1 (>= 1.1.1)
+# debian buster Depends: libc6 (>= 2.14), libglib2.0-0 (>= 2.26.0), liblzma5 (>= 5.1.1alpha+20120614), libssl1.1 (>= 1.1.1)
```
lluiscampos, line 13: Open question, what should we do with old recipes?

As it is `mender-deb-package` now it will always expect these commended lines for armhf builds, but that is not true on `recipes != master` so bugfix releases might get broken Dependencies.

Should we port this back to all recipes or just check in for existence of these lines in `mender-deb-package` before post-processing them (calling `sed`)

docker-build-image
lluiscampos: Thanks, I will do this, much cleaner! :+1:

mender-deb-package
```diff
@@ -101,19 +115,26 @@ prepare_recipe() {
...
-  fi
   dch --create \
     --newversion ${DEB_VERSION} \
-    --distribution ${distribution} \
+    --distribution ${DEB_DISTRIBUTION} \
     --package ${DEB_PACKAGE} \
     "Release ${DEB_VERSION}. See online docs for complete changelog"
 }
 
+postprocess_recipe() {
```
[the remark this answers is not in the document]
lluiscampos: What do you mean? To do a relation like `${shlibs:Depends}, ${misc:Depends} | <hardcoded dependencies for armhf here>`? (that exactly won't work, but it is a high level idea). Can you elaborate a bit more?

recipes/mender-client/debian-master/control
lluiscampos: This we don't need to solve "today", we only need to solve it before we build the bugfix releases (some time next week).

mender-deb-package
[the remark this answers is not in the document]
oleorhagen: Yeh, this is a dummy example, but we could do:
```Depends: ${shlibs:Depends}, ${misc:Depends}, libc6 (>= 2.14), libglib2.0-0 (>= 2.26.0) [armhf], libglib2.0-0 [!armhf], liblzma5 (>= 5.1.1alpha+20120614), libssl1.2 (>= 1.1.1) [!armhf], libssl1.1 (>= 1.1.1) [armhf]
```
for armhf specifics.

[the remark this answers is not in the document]
lluiscampos: Mmm. I still don't get it. If any we would like to have only armhf dependencies, and let ${shlibs:Depends} expand on other architectures. For example:
```
Depends: ${shlibs:Depends}, ${misc:Depends}, libglib2.0-0 (>= 2.26.0) [armhf], libssl1.1 (>= 1.1.1) [armhf]
```

But then, we would need a separate recipe for Ubuntu platforms, where we are using proper Debian toolchain for armhf also.

Overall it is a hacky process. During this rework I was thinking that we should have prepared separate Dockerfile + wrapper script + recipes for Debian armhf, which is the special case, and then the simpler Dockerfile + wrapper sript + recipes for the rest. But yeah, I didn't dare to go that far.

recipes/mender-monitor/debian-master/control
```diff
@@ -15,5 +15,5 @@ Description: Mender Monitor client
  
 Package: mender-monitor-demo
 Architecture: all
-Depends: mender-monitor, ${misc:Depends}
+Depends: mender-monitor (= ${binary:Version}), ${misc:Depends}
```
tranchitella, line 18: Smart!

tranchitella (approved):
Looks good

PR #181, opened 2022-02-01
install-mender.sh: Fix typo in deprecated repo check

lluiscampos:


tranchitella (approved):

scripts/install-mender.sh
```diff
@@ -210,7 +210,7 @@ add_repo() {
         exit 1
     fi
     if test -f /etc/apt/sources.list.d/mender.list && \
```
merlin-northern, line 212:
```
MENDER_APT_SOURCE="/etc/apt/sources.list.d/mender.list"
```
wouldn't that be nice?

as a review I actually double checked that `"/etc/apt/sources.list.d/mender.list"` is actually created by the get.medner.io script.

lluiscampos: okey

scripts/install-mender.sh
```diff
@@ -209,10 +212,10 @@ add_repo() {
         echo "See https://docs.mender.io for updated APT repos information"
         exit 1
     fi
-    if test -f /etc/apt/sources.list.d/mender.list && \
-            grep -F "$repo_deprecated" /etc/apt/sources.list.d/mender.list >/dev/null; then
+    if test -f $MENDER_APT_SOURCES_LIST && \
```
merlin-northern, line 215: sorry I am being really annoying, we need double quotes around all refs. pre-approving. and thank you for applying all my comments!

merlin-northern (approved):

scripts/install-mender.sh
lluiscampos: But why? We are defining this variable in the same script, we know exactly its value. Why do we need the quotes?

merlin-northern (approved):
thank you and sorry for being picky, and annoying

scripts/install-mender.sh
merlin-northern: hypothetical scenario: someone comes and for some reason changes this variable to "/my crazy dir/because" and we can be in trouble. this highly hypothetical person would have to check all the references, which she or he would probably not do.

PR #182, opened 2022-02-02
mender-deb-package: Fix detection of distribution based on tags

lluiscampos:
The logic was wrong: we only want in stable final tags, build tags and
master versions go to experimental.

Amends commit 7530e57.

### DIFF
--- a/mender-deb-package
+++ b/mender-deb-package
@@ -56,7 +56,7 @@ get_os_version() {
 }
 
 get_deb_distribution() {
-  if [ "$VERSION" != "master" ] && git describe --tags --exact-match 2>/dev/null; then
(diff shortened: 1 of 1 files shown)


mender-deb-package
```diff
@@ -56,7 +56,7 @@ get_os_version() {
 }
 
 get_deb_distribution() {
-  if [ "$VERSION" != "master" ] && git describe --tags --exact-match 2>/dev/null; then
+  if [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
oleorhagen, line 59: :+1:

oleorhagen (approved):

PR #192, opened 2022-02-07
pipeline: publish:s3:docs-link:mender-client: Recreate master version

lluiscampos:
For the "master" copy of the package, respect the new package naming by
recreating the correct suffix.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -434,9 +434,9 @@ publish:s3:apt-repo:automatic:
           --key ${MENDER_VERSION}/${S3_BUCKET_SUBPATH}/${arch}/mender-client_${deb_version}_${arch}.deb;
         if [ "${MENDER_VERSION}" == "master" ]; then
           aws s3 cp output/opensource/mender-client_${deb_version}_${arch}.deb
-            s3://${S3_BUCKET_NAME}/${MENDER_VERSION}/${S3_BUCKET_SUBPATH}/${arch}/mender-client_master-1_${arch}.deb;
+            s3://${S3_BUCKET_NAME}/${MENDER_VERSION}/${S3_BUCKET_SUBPATH}/${arch}/mender-client_master-1+debian+buster_${arch}.deb;
           aws s3api put-object-acl --acl public-read --bucket ${S3_BUCKET_NAME}
-            --key ${MENDER_VERSION}/${S3_BUCKET_SUBPATH}/${arch}/mender-client_master-1_${arch}.deb;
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -434,9 +434,9 @@ publish:s3:apt-repo:automatic:
           --key ${MENDER_VERSION}/${S3_BUCKET_SUBPATH}/${arch}/mender-client_${deb_version}_${arch}.deb;
         if [ "${MENDER_VERSION}" == "master" ]; then
           aws s3 cp output/opensource/mender-client_${deb_version}_${arch}.deb
-            s3://${S3_BUCKET_NAME}/${MENDER_VERSION}/${S3_BUCKET_SUBPATH}/${arch}/mender-client_master-1_${arch}.deb;
+            s3://${S3_BUCKET_NAME}/${MENDER_VERSION}/${S3_BUCKET_SUBPATH}/${arch}/mender-client_master-1+debian+buster_${arch}.deb;
```
merlin-northern, line 437: dont we have the distribution flavours in `${deb_version}` ?

lluiscampos: Yes, that works just fine for tagged versions in lines above. But here we do a hardcoded copy for this special link to master pacakges in the docs.

We could extract "debian+buster" from `${deb_version}` but that is really an overkill IMO when we know exactly which packages are we copying (see lines 420-423)

merlin-northern: if you say so. then it is ok. (I think we should use deb_version in both src and dst of aws command, or hardcode in both places, otherwise it does not look right for me)

merlin-northern (approved):
one comment, up to you!

.gitlab-ci.yml
lluiscampos: I am sorry I am not able to explain it correctly. I'll try again.

We cannot use `deb_version` both in src and dst because it is not the same. What these lines are doing is making a "dirty copy" of a file named, for example, `mender-client_3.3.0~git20220201.e005a64-1+debian+buster+b465041804_amd64.deb` as `mender-client_master-1+debian+buster+b465041804_amd64.deb`. In this example `deb_version` is `3.3.0~git20220201.e005a64-1+debian+buster+b465041804` and the only reusable part of that is the `+debian+buster` suffix for which I commented I think it is an overkill to do some sed magic just to extract that.

The copy where src and dst are the same are few lines above, out of this special `if` condition that only applies for master packages (which also applies for `master` packages, of course)

PR #197, opened 2022-02-07
MEN-5405 Debian Source packages.

merlin-northern:
ChangeLog:title

### DIFF
--- a/docker-build-package
+++ b/docker-build-package
@@ -99,7 +99,6 @@ if [ "$arch_indep" = "true" -a "$ARCH" != "amd64" ]; then
 fi
 
 echo "Building $recipe_name..."
-echo ""
 
 output_dir="output/opensource"
 if [ $commercial = "true" ]; then
@@ -117,6 +116,12 @@ else
     # On others, build only architecture dependent packages.
     BUILD_TYPE=any
 fi
+
+if [ "$commercial" != "true" ]; then
+    echo "Including source packages in the build."
+    BUILD_TYPE="source,${BUILD_TYPE}"
+fi
+echo
 
 docker run --rm \
         --volume $(pwd)/recipes:/recipes \
--- a/mender-deb-package
+++ b/mender-deb-package
@@ -70,7 +70,9 @@ get_deb_version() {
   #     where X.Y.Z is latest tag (not necessarily matching git describe)
   debian_suffix="1+$OS_DISTRO+$OS_CODENAME"
   if [ "$VERSION" != "master" ] && git describe --tags --exact-match 2>/dev/null; then
-    DEB_VERSION="$(git describe --tags --exact-match)-$debian_suffix"
+    DEB_VERSION="$(git describe --tags --exact-match)"
+    DEB_VERSION_NO_REV="${DEB_VERSION}"
+    DEB_VERSION="${DEB_VERSION}-${debian_suffix}"
   else
     DEB_VERSION="$(git tag | egrep '^[0-9]+\.[0-9]+\.[0-9]+$' | sort -rV | head -n1)"
     if [ -z "${DEB_VERSION}" ]; then
@@ -80,10 +82,15 @@ get_deb_version() {
     DEB_VERSION="$(echo ${DEB_VERSION} | sed -E 's/([0-9]+\.)([0-9]+)\.[0-9]+/echo "\1$((\2+1))\.0"/e')"
     # Append git date and commit
     DEB_VERSION="${DEB_VERSION}$(git log -1 --pretty=~git%cd.%h --date format:%Y%m%d)"
+    DEB_VERSION_NO_REV="${DEB_VERSION}"
     # Append Debian suffix
     DEB_VERSION="${DEB_VERSION}-$debian_suffix"
     # Append build number
-    DEB_VERSION="${DEB_VERSION}+b${BUILD_ID}"
+    # there is a bizarre feature: it can't be `bNUMBER`
+    # or the source won't build (will not be able to find dsc file)
+    # this has to have something to do with the `b` being a prefix
+    # to binary numbers in perl
+    DEB_VERSION="${DEB_VERSION}+builder${BUILD_ID}"
   fi
 }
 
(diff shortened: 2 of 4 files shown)


mender-deb-package
```diff
@@ -80,10 +80,15 @@ get_deb_version() {
     DEB_VERSION="$(echo ${DEB_VERSION} | sed -E 's/([0-9]+\.)([0-9]+)\.[0-9]+/echo "\1$((\2+1))\.0"/e')"
     # Append git date and commit
     DEB_VERSION="${DEB_VERSION}$(git log -1 --pretty=~git%cd.%h --date format:%Y%m%d)"
+    DEB_VERSION_NO_REV="${DEB_VERSION}"
```
lluiscampos, line 83: `DEB_VERSION_NO_REV` needs to be defined also on tags, line 73 above

mender-deb-package
```diff
@@ -149,12 +154,18 @@ build_packages() {
     sign_flags="--sign-key=$key_id"
   fi
 
+  # dh_make needs a name of the user in the env
+  export USER=root
+  # needed to get the orig file
+  echo dh_make --createorig --packagename "${DEB_PACKAGE}_${DEB_VERSION_NO_REV}" -s -y || true
```
lluiscampos, line 160: debug? Please remove the echo - the whole script is executed with `set -x` anyway

mender-deb-package
```diff
@@ -80,10 +80,15 @@ get_deb_version() {
...
     DEB_VERSION="${DEB_VERSION}$(git log -1 --pretty=~git%cd.%h --date format:%Y%m%d)"
+    DEB_VERSION_NO_REV="${DEB_VERSION}"
     # Append Debian suffix
     DEB_VERSION="${DEB_VERSION}-$debian_suffix"
     # Append build number
-    DEB_VERSION="${DEB_VERSION}+b${BUILD_ID}"
+    # there is a bizarre feature: it can't be `bNUMBER`
+    # or the source won't build (will not be able to find dsc file)
+    # this has to have something to do with the `b` being a prefix
+    # to binary numbers in perl
```
lluiscampos, line 90: Great finding :wink:

mender-deb-package
```diff
@@ -149,12 +154,18 @@ build_packages() {
     sign_flags="--sign-key=$key_id"
   fi
 
+  # dh_make needs a name of the user in the env
+  export USER=root
+  # needed to get the orig file
+  echo dh_make --createorig --packagename "${DEB_PACKAGE}_${DEB_VERSION_NO_REV}" -s -y || true
+  dh_make --createorig --packagename "${DEB_PACKAGE}_${DEB_VERSION_NO_REV}" -s -y || true
```
lluiscampos, line 161: Why the `|| true` ? Shouldn't we make the script fail if the orig cannot be created?

lluiscampos (changes requested):
The design looks good. It needs a fix for building Git tags and I also wrote some other minor comments.

Please launch a build using existing tags (maybe one final tag and other `-buildX` tag) to verify that the script works fine then.

mender-deb-package
merlin-northern: the problem here is: dh_make will exit with 1 because debian directory exists -- but do not worry, if the orig file is not created here, the next command fails -- so no it should be ok.

mender-deb-package
merlin-northern: debug, I will remove.

mender-deb-package
merlin-northern: ah right! I thought this line was covered with tests. thank you.

lluiscampos: There are no tests for this script

mender-deb-package
lluiscampos: ok

lluiscampos (approved):
Looks good to me now!

> Please launch a build using existing tags (maybe one final tag and other -buildX tag) to verify that the script works fine then.

Remember to try this before merging.

mender-deb-package
```diff
@@ -149,12 +156,18 @@ build_packages() {
...
+  # hence the || true, without orig file buildpackage command would fail
+  dh_make --createorig --packagename "${DEB_PACKAGE}_${DEB_VERSION_NO_REV}" -s -y || true
+
   case "$ARCH" in
     amd64)
       # Native build (amd64)
       dpkg-buildpackage \
         ${sign_flags} \
-        --build=$DEB_BUILD_TYPE
+        --build=source,$DEB_BUILD_TYPE
```
lluiscampos, line 170: This will create source tarballs also for commercial packages.

Although we can then handle it in the publish stem to make sure we don't publish them anywhere... I think is best to just not build them and make sure this way that we don't publish them by mistake.

With that in mind, here we should revert to use `$DEB_BUILD_TYPE`, and in [here](https://github.com/mendersoftware/mender-dist-packages/blob/master/docker-build-package#L120) prepend `source,` to the variable for non commercial packages.

What do you think?

lluiscampos (changes requested):

mender-deb-package
merlin-northern: makes sense. thank you. fixed.

docker-build-package
```diff
@@ -118,6 +118,12 @@ else
     BUILD_TYPE=any
 fi
 
+if [ "$commercial" != "true" ]; then
+    echo "Including source packages in the build."
+    echo
```
lluiscampos, line 123: Nitpick: suggestion to remove this `echo` and the one in line `102` and add instead a single one on line `126` after this if condition.

lluiscampos (approved):

docker-build-package
merlin-northern: [sure](https://github.com/mendersoftware/mender-dist-packages/compare/7318ac82ef072107642e29220fd32596325f7ca3..c7b535bf6c0c2e7f2e7246db84f290930bbe4296)

merlin-northern:
> Looks good to me now!
> 
> > Please launch a build using existing tags (maybe one final tag and other -buildX tag) to verify that the script works fine then.
> 
> Remember to try this before merging.

https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/466936247
 but I think I have to do something more to build commercial

PR #198, opened 2022-02-08
pipeline: Move build jobs to private runners

lluiscampos:


oleorhagen (approved):
:+1:

lluiscampos:
Note to self,

It does not work when triggering this pipeline from `mender-qa` due to the Docker in Docker settings. Review this.

PR #206, opened 2022-02-18
Fixes to debian.tar.xz files generation.

merlin-northern:
This one addresses the problem of checksums mismatch for debian.tar.xz files, which are generated per distribution,
e.g.: `mender-client_3.3.0~git20220215.a905a0a-1+debian+buster+builder474683108.debian.tar.xz`, but contain 
the timestamp of generation and the username and user email in the debian/changelog. It ensures that the debian/changelog will be the same across all the architectures.

The other option is to create source packages per distribution, right now we are building sources for all cases.

ChangeLog:title

### DIFF
--- a/docker-build-package
+++ b/docker-build-package
@@ -117,7 +117,7 @@ else
     BUILD_TYPE=any
 fi
 
-if [ "$commercial" != "true" ]; then
+if [ "$commercial" != "true" -a "$ARCH" = "amd64" ]; then
     echo "Including source packages in the build."
     BUILD_TYPE="source,${BUILD_TYPE}"
 fi


mender-deb-package
```diff
@@ -2,6 +2,9 @@
 
 set -ex
 
+export DEBEMAIL=${DEBEMAIL:-"Mender root"}
+export DEBFULLNAME=${DEBFULLNAME:-"root@mender.io"}
+
```
merlin-northern, line 7: we were generating the `debian/changelog` file (below with dch) with root@<random-hostname-based-on-docker-container>, however devian/changelog is packaged into debian.tar.xz files, which are in turn added to dsc files, with a checksum. Now the checksum was different every time, but the debian.tar.xz files are generated per-deistribution (not per architecture and distribution): e.g.: `mender-client_3.3.0~git20220215.a905a0a-1+debian+buster+builder474683108.debian.tar`, that in turn lead to mismatched checksums while importing some packages:
```
ERROR: Unexpected content of file '/var/reprepro/incoming/mender-client_3.3.0~git20220215.a905a0a-1+debian+bullseye+builder474191641.debian.tar.xz'!
md5 expected: f80e29b7922deaef3768eaf695b682d9, got: 3083936e29d6801603d185c46afccb94
sha1 expected: 350c55457a8099568997ec55937cc54d64455b69, got: 50d7a6e57afdd178b93ca89e9db7c08f70438522
sha256 expected: 7417f752d4b841339a928650e8f1d8f664a8fae36e0405177181281cef772066, got: 9aafb01667f786f4c13a1b1819bf1635b536b05c449aa0018b27ab3b80ccb432
size expected: 5332, got: 5348
```
(it depended on the order of builds finished)

mender-deb-package
```diff
@@ -122,7 +127,9 @@ prepare_recipe() {
     cp support/mender-monitor.service debian/
   fi
 
-  dch --create \
+
+  timestamp="`date +'%a %b %d 12:12:00 %Z %Y'`"
+  faketime "${timestamp}" dch --create \
```
merlin-northern, line 132: `debian/changelog` contains also date, for the same reasons as above: https://github.com/mendersoftware/mender-dist-packages/pull/206/commits/628d55388454e484367d9fe5567c6efe8327228e#r810799323 we cannot generate a changelog entry with the current timestamp.

oleorhagen (commented):
I'm going to defer this one to @lluiscampos

lluiscampos (changes requested):
I understand what you are trying to solve but I think the approach here is an overkill.

We had a similar problem when building architecture independent packages on different architectures which resulted in different checksums, and the solution was to build these only on one architecture (we decided upon amd64). As a side effect, it became more efficient.

I suggest we do the same, to only build the source packages on amd64. Basically extending [this if condition](https://github.com/mendersoftware/mender-dist-packages/blob/master/docker-build-package#L120) with `-a "$ARCH" = "amd64" `.

merlin-northern:
> I understand what you are trying to solve but I think the approach here is an overkill.
> 
> We had a similar problem when building architecture independent packages on different architectures which resulted in different checksums, and the solution was to build these only on one architecture (we decided upon amd64). As a side effect, it became more efficient.
> 
> I suggest we do the same, to only build the source packages on amd64. Basically extending [this if condition](https://github.com/mendersoftware/mender-dist-packages/blob/master/docker-build-package#L120) with `-a "$ARCH" = "amd64" `.

makes sense! (that was the other option I mentioned called "by distribution")

lluiscampos:
> makes sense! (that was the other option I mentioned called "by distribution")

Right, but it is important to distinguish that what I mean is "by architecture" not "by distribution" - we will still build them for all distributions but then it is fine that they have different checksums because they are different files anyway.

merlin-northern:
@lluiscampos  thank you! also: I hate when I debug something for hours and you come and fix it just-in-passing in 15seconds and with a couple of bytes worth of diff. :)

PR #213, opened 2022-03-06
MEN-5460 + MEN-5029: Commercial packages publish fixes

lluiscampos:


.gitlab-ci.yml
```diff
@@ -549,3 +542,49 @@ publish:s3:mender-gateway:automatic:
...
+    - build:packages:ubuntu:bionic:arm64
+    - build:packages:ubuntu:focal:amd64
+    - build:packages:ubuntu:focal:armhf
+    - build:packages:ubuntu:focal:arm64
+  before_script:
+    - apt update && apt install -yyq awscli
+    - *publish_helper_functions
+  script:
+    - echo "Publishing "latest" package for ${PUBLISH_PACKAGE_PREFIX} to s3://${S3_BUCKET_NAME_PRIVATE}/${PUBLISH_PACKAGE_S3_SUBPATH}/${PUBLISH_PACKAGE_VERSION}/"
+    - if ! is_final_tag ${PUBLISH_PACKAGE_VERSION}; then
```
merlin-northern, line 568: paranoia mode on: I would double quote the version here.

.gitlab-ci.yml
```diff
@@ -482,33 +482,34 @@ publish:production:s3:scripts:install-mender-sh:
...
     - build:packages:debian:buster:arm64
+    - build:packages:debian:bullseye:amd64
+    - build:packages:debian:bullseye:armhf
+    - build:packages:debian:bullseye:arm64
+    - build:packages:ubuntu:bionic:amd64
+    - build:packages:ubuntu:bionic:armhf
+    - build:packages:ubuntu:bionic:arm64
+    - build:packages:ubuntu:focal:amd64
+    - build:packages:ubuntu:focal:armhf
+    - build:packages:ubuntu:focal:arm64
```
merlin-northern, line 500: do we need to change this: [[arch_indep]="true"](https://github.com/mendersoftware/mender-dist-packages/blob/master/docker-build-package#L42) as well?

merlin-northern (commented):
not an expert here, left come comments.

.gitlab-ci.yml
lluiscampos: Why? I don't get it.

It is true that for `mender-monitor` all the `:arm*` jobs won't have valuable artifacts, but it is harmless to fetch them. And this way we are able to use the same template to all commercial packages and avoid duplicating code.

.gitlab-ci.yml
lluiscampos: Sure, I'll do that.

scripts/install-mender.sh
```diff
@@ -443,6 +436,10 @@ check_dist_and_version() {
 
     echo "  Detected distribution:"
     printf "\t%s/%s\n" "$LSB_DIST" "$DIST_VERSION"
+
+    if [[ "$LSB_DIST" == "raspbian" ]]; then
```
oleorhagen, line 440: Suggestion:

Just a short log message here probably won't hurt us.

.gitlab-ci.yml
```diff
@@ -549,3 +542,49 @@ publish:s3:mender-gateway:automatic:
...
+    -   echo "Attempting to publish a "latest" package for a non final tag!"
+    -   exit 1
+    - fi
+    - find output/commercial -name "${PUBLISH_PACKAGE_PREFIX}*.deb" | while read -r file; do
+    #   Make copy for "latest" to be consumed by get.mender.io script:
+    -   copy_filename=$(basename "$file" | sed 's/_[^\+]*/_latest-1/; s/\+builder[^_]*//')
+    -   aws s3 cp "$file" s3://${S3_BUCKET_NAME_PRIVATE}/${PUBLISH_PACKAGE_S3_SUBPATH}/${PUBLISH_PACKAGE_VERSION}/${copy_filename}
+    - done
+
+publish:s3:mender-monitor:latest:
```
oleorhagen, line 578: Suggestion:

Name them manual as well?

oleorhagen (approved):
Nice !

Left a few nitties, feel free to ignore at will

scripts/install-mender.sh
lluiscampos: I went a bit back and forth with this because I couldn't find a good way of logging it. But once you comment on it, means that I have to do a better effort :) I'll try

.gitlab-ci.yml
lluiscampos: I will. Thanks :+1:

.gitlab-ci.yml
merlin-northern: > Why?

the only why was: because we are building monitor debs only for amd64.

> 
> It is true that for `mender-monitor` all the `:arm*` jobs won't have valuable artifacts, but it is harmless to fetch them. And this way we are able to use the same template to all commercial packages and avoid duplicating code.

makes sense. I also mis-read, and thoughr we are now building monitor per arch as well. sorry for the noise.

PR #215, opened 2022-03-07
test_package_client.py: Fix full cycle detection

lluiscampos:
After modification in the client state machine. See:
https://github.com/mendersoftware/mender/pull/943

Add a new message to the criterion for the first state machine
transition and give more time to get to the authorization failure
(because now bootstrap keys get generated _after_ the transition to
inventory-update state).

### DIFF
--- a/tests/test_package_client.py
+++ b/tests/test_package_client.py
@@ -1,5 +1,5 @@
 #!/usr/bin/python3
-# Copyright 2021 Northern.tech AS
+# Copyright 2022 Northern.tech AS
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
 #    you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


oleorhagen (approved):

tests/test_package_client.py
```diff
@@ -116,15 +118,23 @@ def check_systemd_start_full_cycle(self, ssh_connection):
         assert "Loaded configuration file" in result.stdout, result.stdout
         assert "No dual rootfs configuration present" in result.stdout, result.stdout
 
-        assert any(
-            [
-                expected_output in result.stdout
-                for expected_output in [
-                    "Reauthorization failed with error: transient error: authorization request failed",
-                    "Authorize failed: transient error: authorization request failed",
+        while time.time() - start_time < timeout:
```
merlin-northern, line 121: just wondering if `start_time = time.time()` should be set here?

merlin-northern (approved):
lgtm with one comment

tests/test_package_client.py
lluiscampos: No. The timeout for the whole cycle are these 4 minutes that we set at the beginning.

PR #216, opened 2022-03-07
MEN-5475: install-mender.sh: Support setting up mender-gateway for demo

lluiscampos:
When supplied with `--demo` flag, the script will download and extract
the examples tarball of mender-gateway containing configuration file,
certificate and key for demo purposes.

### DIFF
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -42,16 +42,19 @@ DEMO="0"
 # Path where to install the Mender APT repository
 MENDER_APT_SOURCES_LIST="/etc/apt/sources.list.d/mender.list"
 
-# URL prefix from where to download commercial compoments
+# URL prefix for the commercial components
 MENDER_COMMERCIAL_DOWNLOAD_URL="https://downloads.customer.mender.io/content/hosted/"
 
-# URL path for the actual components, formatted by version
+# URL path for the commercial components, formatted by version, distribution and release
 ARCHITECTURE=$(dpkg --print-architecture)
 declare -A COMMERCIAL_COMP_TO_URL_PATH_F=(
   [mender-gateway]="mender-gateway/debian/%s/mender-gateway_%s-1+%s+%s_$ARCHITECTURE.deb"
   [mender-monitor]="mender-monitor/debian/%s/mender-monitor_%s-1+%s+%s_all.deb"
   [mender-monitor-demo]="mender-monitor/debian/%s/mender-monitor-demo_%s-1+%s+%s_all.deb"
 )
+
+# URL path for mender-gateway demo, formatted by version
+MENDER_GATEWAY_EXAMPLES_URL_PATH_F="mender-gateway/examples/%s/mender-gateway-examples-%s.tar"
 
 export DEBIAN_FRONTEND=noninteractive
 
@@ -195,6 +198,15 @@ init() {
 
     echo "  Installing from channel:"
     printf "\t%s\n" "$CHANNEL"
+
+    # Translate Debian "channel" into Mender version for commercial packages
+    if [ "$CHANNEL" = "experimental" ]; then
+        VERSION="master"
+    else
+        VERSION="latest"
+    fi
+    echo "  Installing commercial components from:"
+    printf "\t%s\n" "$VERSION"
 }
 
 get_deps() {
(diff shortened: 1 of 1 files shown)


lluiscampos:
These go together:
* https://github.com/mendersoftware/mender-gateway/pull/17
* https://github.com/mendersoftware/mender-qa/pull/572
* https://github.com/mendersoftware/meta-mender/pull/1613
* https://github.com/mendersoftware/mender-dist-packages/pull/216

scripts/install-mender.sh
```diff
@@ -270,16 +283,9 @@ do_install_commercial() {
...
-    if [ "$CHANNEL" = "experimental" ]; then
-        version="master"
-    else
-        version="latest"
-    fi
-
     # Download deb packages
     for c in $selected_components_commercial; do
-        url="$MENDER_COMMERCIAL_DOWNLOAD_URL$(printf ${COMMERCIAL_COMP_TO_URL_PATH_F[$c]} $version $version $LSB_DIST $DIST_VERSION)"
+        url="$MENDER_COMMERCIAL_DOWNLOAD_URL$(printf ${COMMERCIAL_COMP_TO_URL_PATH_F[$c]} $VERSION $VERSION $LSB_DIST $DIST_VERSION)"
```
merlin-northern, line 288: * c should be local
* url should be local
* $MENDER_COMMERCIAL_DOWNLOAD_URL -> `${MENDER_COMMERCIAL_DOWNLOAD_URL}`
*

scripts/install-mender.sh
```diff
@@ -328,6 +334,23 @@ EOF
             echo "  Success!"
         fi
     fi
+
+    # Setup for mender-gateway
+    if [[ "$SELECTED_COMPONENTS" == *"mender-gateway"* ]]; then
+        if [ "$DEMO" -eq 1 ]; then
+            echo "  Setting up mender-gateway with demo configuration, certificates and key"
+            url="$MENDER_COMMERCIAL_DOWNLOAD_URL$(printf ${MENDER_GATEWAY_EXAMPLES_URL_PATH_F} $VERSION $VERSION)"
```
merlin-northern, line 342: url should be local

scripts/install-mender.sh
```diff
@@ -328,6 +334,23 @@ EOF
...
+            curl -fLsS -H "Authorization: Bearer $JWT_TOKEN" -O "$url" ||
+                    (echo ERROR: Cannot get mender-gateway-examples from $url; exit 1)
+            extract_dir=$(mktemp -d)
+            tar -C $extract_dir -xf mender-gateway-examples-$VERSION.tar
+            rsync --links --recursive $extract_dir/mender-gateway-examples-$VERSION/* /
```
merlin-northern, line 347: * apart from double quotes around `${extract_dir}` I would also use `${VERSION}`
* this is about decompressing to / with trimming of the `mender-gateway-examples-$VERSION/` in leading paths, right? is so, then: `tar -C / --strip-components=1 -xf mender-gateway-examples-$VERSION.tar` (but please test it typing from half-memory)

scripts/install-mender.sh
```diff
@@ -328,6 +334,23 @@ EOF
...
         fi
     fi
+
+    # Setup for mender-gateway
+    if [[ "$SELECTED_COMPONENTS" == *"mender-gateway"* ]]; then
+        if [ "$DEMO" -eq 1 ]; then
+            echo "  Setting up mender-gateway with demo configuration, certificates and key"
+            url="$MENDER_COMMERCIAL_DOWNLOAD_URL$(printf ${MENDER_GATEWAY_EXAMPLES_URL_PATH_F} $VERSION $VERSION)"
+            curl -fLsS -H "Authorization: Bearer $JWT_TOKEN" -O "$url" ||
+                    (echo ERROR: Cannot get mender-gateway-examples from $url; exit 1)
```
merlin-northern, line 344: * it will never contain URL encoded characters? (see the wget excursion we have enjoyed so much earlier)
* the exit here will only exit the subshell, is that on purpose?

scripts/install-mender.sh
lluiscampos: @merlin-northern I didn't know about `--strip-components=1`, thank you very much! :pray:

scripts/install-mender.sh
lluiscampos: I moved it into an explicit `if` function

tranchitella (approved):
Looks good, I checked the path to the files in S3 and it matches.

oleorhagen (approved):
Nice!

scripts/install-mender.sh
merlin-northern: the if looks good. another option was to use the group command: `... || {...; exit 1; }` but it is fine.
the wget/curl fun was due to curl not de-encoding the URL-encoded filenames, but if you have it under control then it is ok.

merlin-northern (approved):

scripts/install-mender.sh
```diff
@@ -195,6 +198,15 @@ init() {
...
     echo "  Installing from channel:"
     printf "\t%s\n" "$CHANNEL"
+
+    # Translate Debian "channel" into Mender version for commercial packages
+    if [ "$CHANNEL" = "experimental" ]; then
+        VERSION="master"
+    else
+        VERSION="latest"
+    fi
+    echo "  Installing commercial components from:"
```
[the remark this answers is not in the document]
lluiscampos: @oleorhagen Right. I know it is a bit ambiguous but it is not really a "channel", it is something else. If we state here "channel", the UX becomes confusing: the user specifies "channel" `experimental` but gets "channel" `master` (or "channel" `stable` to `latest`).

So, what do you think? Maybe call it "Installing commercial components version:" or "Installing commercial components from source"... ?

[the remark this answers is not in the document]
oleorhagen: The latter sounds appropriate to me.

PR #219, opened 2022-03-14
install-mender.sh: Amend echo message for commercial components

lluiscampos:


lluiscampos:
See https://github.com/mendersoftware/mender-dist-packages/pull/216#discussion_r824555087

PR #229, opened 2022-04-19
Work around failure to resolve foreign arch deps in ubuntu/focal.

kacf:
It is unknown why, but it seems to work to add them manually.


oleorhagen (approved):
No idea what's going on here...

PR #253, opened 2022-07-05
do not package .git or debian directory, publish same orig file

merlin-northern:
exclude .git and debian directories in orig file, use one orig file.

 * build the orig.tar file by ourselves: excluding .git (not needed) and
   debian (present in debian.tar file) directories
 * tar produces different checksums for identical directories occupying
   different inodes: in the publish stage we copy one and same orig file
   (the contents of it never change between distros, release names, archs)

Ticket: QA-404


merlin-northern:
I am checking one thing, ignore for now.

tranchitella (approved):

PR #254, opened 2022-07-05
chore: A more elegant approach of using a reference orig file.

merlin-northern:
Changelog: Title
Ticket: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -39,6 +39,7 @@ include:
 
 stages:
   - build:images
+  - build:orig
   - build:packages
   - test
   - publish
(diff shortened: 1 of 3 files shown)


merlin-northern:
ignore for now please

merlin-northern:
@oleorhagen I do not know of any other nice way to do it, and I really do not like the other solutions I proposed.

.gitlab-ci.yml
```diff
@@ -165,6 +193,7 @@ build:image:ubuntu:focal:arm64:
     - docker load -i builder-$DISTRO-$RELEASE-$ARCH.tar
   script:
     - apk --update --no-cache add bash
+    - find output
```
kacf, line 196: Minor: Isn't this more useful after the building steps?

docker-build-package
```diff
@@ -104,7 +105,9 @@ output_dir="output/opensource/${DISTRO}-${RELEASE}-${ARCH}"
 if [ $commercial = "true" ]; then
     output_dir="output/commercial/${DISTRO}-${RELEASE}-${ARCH}"
 fi
-mkdir -p ${output_dir}
+mkdir -p "${output_dir}"
+find output -name '*.orig.tar.*'
+[[ "${SAVE_ORIG}" != "true" && "${DISTRO}-${RELEASE}-${ARCH}" != "debian-buster-amd64" ]] && cp -v output/opensource/debian-buster-amd64/*orig.tar.* "${output_dir}/"
```
kacf, line 110: This hardcoded path is quite ugly. Suggestion: Copy the orig file to some common place instead, while building it, and then pick that up here instead of a hardcoded platform path.

kacf (approved):
I think the one hardcoded path below should be changed, but otherwise looks good, so I'm pre-approving.

.gitlab-ci.yml
merlin-northern: moved down. it makes sense in both places, before: so we know that the required files are there and after so we know what we have built.

docker-build-package
merlin-northern: do you like [output/opensource/orig](https://github.com/mendersoftware/mender-dist-packages/pull/254/commits/1de010f32915945c1311780818ae0f2c1e01185f#diff-a696dd766c3d4eb6a58ba77a8c75d4f3888cf19cb5d7332f12a6214f07f0d5d1R108) better?

PR #258, opened 2022-07-15
(ignore) More elegant approach review fixes

merlin-northern:


merlin-northern:
thanks @lluiscampos for reminding me to tidy up.

PR #260, opened 2022-07-17
chore: do not call build package in orig stage for commercial packages.

merlin-northern:
Ticket: None


merlin-northern:
@kacf there is no need to call build in orig stage for the commercial packages (we do not build source packages for them).

kacf (approved):

PR #270, opened 2022-09-05
test(test_mender_gateway): Amend commit d840e586

lluiscampos:
This test does not run on PRs, so the mistake went unnoticed on the
review.


lluiscampos:
You can see the test failing [here](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/jobs/2980383684).

tranchitella (approved):

PR #277, opened 2022-09-16
chore: Add comment in pipeline cross-referencing mender-qa trigger

lluiscampos:


lluiscampos:
The motivation is for us not to miss again these kind of updates:
* https://github.com/mendersoftware/mender-qa/pull/609

PR #278, opened 2022-09-22
chore(install-mender.sh): Clean-up also example package after install

lluiscampos:
Amends commit a17b523c7a986f37ccefe4fc7f59030d98e6643c


oleorhagen (approved):
Maybe we should consider moving to a flow where we just store everything to be deleted in an array, and clean up in an exit hook :thinking:

lluiscampos:
> Maybe we should consider moving to a flow where we just store everything to be deleted in an array, and clean up in an exit hook thinking

How it should be done IMO is create a tmpdir and change the cwd there.

I'll merge it as is because it is what I tested during the release, but take note for the further improvement :memo:

PR #281, opened 2022-09-27
feat: Do not publish more orig files than needed

merlin-northern:
The publication of the packages to the apt-repo left some orig files.
It turns out those were unreferenced orig files coming from two sources:

* opensource/orig/ directory: where we store the orig file in ci pipeline
* orig files copied during the binary package build that got propagated
  but never had an entry in the .changes file

We do not need to publish the `orig` directory nor the unreferenced
orig files. This patch takes care of that.

Changelog: Title
Ticket: MEN-5879

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -248,7 +248,6 @@ build:packages:debian:buster:armhf:
     RELEASE: "buster"
     ARCH: "armhf"
   dependencies:
-    - build:orig:debian:buster:amd64
     - build:image:debian:buster:armhf
   extends: .template:build:packages
 
@@ -258,7 +257,6 @@ build:packages:debian:buster:arm64:
     RELEASE: "buster"
     ARCH: "arm64"
   dependencies:
-    - build:orig:debian:buster:amd64
     - build:image:debian:buster:arm64
   extends: .template:build:packages
 
@@ -278,7 +276,6 @@ build:packages:debian:bullseye:armhf:
     RELEASE: "bullseye"
     ARCH: "armhf"
   dependencies:
-    - build:orig:debian:buster:amd64
     - build:image:debian:bullseye:armhf
   extends: .template:build:packages
 
@@ -288,7 +285,6 @@ build:packages:debian:bullseye:arm64:
     RELEASE: "bullseye"
     ARCH: "arm64"
   dependencies:
-    - build:orig:debian:buster:amd64
     - build:image:debian:bullseye:arm64
   extends: .template:build:packages
 
@@ -308,7 +304,6 @@ build:packages:ubuntu:bionic:armhf:
     RELEASE: "bionic"
     ARCH: "armhf"
   dependencies:
-    - build:orig:debian:buster:amd64
     - build:image:ubuntu:bionic:armhf
   extends: .template:build:packages
 
@@ -318,7 +313,6 @@ build:packages:ubuntu:bionic:arm64:
     RELEASE: "bionic"
     ARCH: "arm64"
   dependencies:
-    - build:orig:debian:buster:amd64
     - build:image:ubuntu:bionic:arm64
   extends: .template:build:packages
 
@@ -338,7 +332,6 @@ build:packages:ubuntu:focal:armhf:
     RELEASE: "focal"
     ARCH: "armhf"
   dependencies:
-    - build:orig:debian:buster:amd64
     - build:image:ubuntu:focal:armhf
   extends: .template:build:packages
 
@@ -348,7 +341,6 @@ build:packages:ubuntu:focal:arm64:
     RELEASE: "focal"
     ARCH: "arm64"
   dependencies:
-    - build:orig:debian:buster:amd64
     - build:image:ubuntu:focal:arm64
   extends: .template:build:packages
 
@@ -368,7 +360,6 @@ build:packages:ubuntu:jammy:armhf:
     RELEASE: "jammy"
     ARCH: "armhf"
   dependencies:
-    - build:orig:debian:buster:amd64
     - build:image:ubuntu:jammy:armhf
   extends: .template:build:packages
 
@@ -378,7 +369,6 @@ build:packages:ubuntu:jammy:arm64:
     RELEASE: "jammy"
     ARCH: "arm64"
   dependencies:
-    - build:orig:debian:buster:amd64
     - build:image:ubuntu:jammy:arm64
   extends: .template:build:packages
 
--- a/docker-build-package
+++ b/docker-build-package
@@ -116,7 +116,12 @@ elif [ "${SAVE_ORIG}" == "true" ]; then
 fi
 mkdir -p "${output_dir}"
 
-[[ "${SAVE_ORIG}" != "true" ]] && cp -v output/opensource/orig/*orig.tar.* "${output_dir}/"
+# we build source packages only for amd64 and for non-commercial
+# (non source closed) components. there is no need to copy the orig
+# file if we do not need it, since later, during publication to the apt-repo
+# we clean them from incoming based on changes file, which refers to orig files
+# only for source builds. See MEN-5879.
+[[ "${SAVE_ORIG}" != "true" && "$commercial" != "true" && "$ARCH" == "amd64" ]] && mv -v output/opensource/orig/${recipe_name}*orig.tar.* "${output_dir}/"
 
 if [ "$arch_indep" = "true" -a "$ARCH" = "amd64" ]; then
     # On amd64, build architecture independent packages.


.gitlab-ci.yml
```diff
@@ -506,7 +506,7 @@ test:acceptance:
   script:
     # Upload files: first .buildinfo and .deb, then .changes
     - for d in output/opensource/*; do
-    -   find "$d" -name '*.buildinfo' -o -name '*.deb' -o -name '*.orig.tar.*' -o -name '*.debian.tar.*' -o -name '*.dsc' | while read -r file; do
+    -   find "$d" \( -name '*.buildinfo' -o -name '*.deb' -o -name '*.orig.tar.*' -o -name '*.debian.tar.*' -o -name '*.dsc' \) -and -not -path "${d}/orig/*" | while read -r file; do
```
lluiscampos, line 509: By design, this is still wrong. The publish job should not need to filter `/output/opensource/orig/` because these are not artifacts of any of the `build:packages:<distro>:<version>:<arch>` jobs, they are artifacts from ``build:orig:debian:buster:amd64` which the publish job does not depend on.

At the high level, this is what should happen:
* `build:orig:debian:buster:amd64 ` produces `output/opensource/orig/*.orig.tar.gz`
* `build:packages:<distro>:<version>:amd64` produces `/output/opensource/<distro>-<version>-amd64/*.orig.tar.gz` but does **not** propagate `output/opensource/orig/*.orig.tar.gz`.

Does it make sense? For example let's take `mender-artifact` files from the build in this PR:
```
./job-build-debian-arm64/output/opensource/orig/mender-artifact_3.10.0~git20220926.4a5b4a0.orig.tar.gz
./job-build-debian-arm64/output/opensource/debian-bullseye-arm64/mender-artifact_3.10.0~git20220926.4a5b4a0-1+debian+bullseye+builder652559327_arm64.deb
./job-build-debian-arm64/output/opensource/debian-bullseye-arm64/mender-artifact_3.10.0~git20220926.4a5b4a0-1+debian+bullseye+builder652559327_arm64.buildinfo
./job-build-debian-arm64/output/opensource/debian-bullseye-arm64/mender-artifact_3.10.0~git20220926.4a5b4a0-1+debian+bullseye+builder652559327_arm64.changes
./job-build-debian-arm64/output/opensource/debian-bullseye-arm64/mender-artifact-deb-version
./job-build-debian-amd64/output/opensource/orig/mender-artifact_3.10.0~git20220926.4a5b4a0.orig.tar.gz
./job-build-debian-amd64/output/opensource/debian-bullseye-amd64/mender-artifact_3.10.0~git20220926.4a5b4a0-1+debian+bullseye+builder652559327.debian.tar.xz
./job-build-debian-amd64/output/opensource/debian-bullseye-amd64/mender-artifact_3.10.0~git20220926.4a5b4a0.orig.tar.gz
./job-build-debian-amd64/output/opensource/debian-bullseye-amd64/mender-artifact_3.10.0~git20220926.4a5b4a0-1+debian+bullseye+builder652559327.dsc
./job-build-debian-amd64/output/opensource/debian-bullseye-amd64/mender-artifact_3.10.0~git20220926.4a5b4a0-1+debian+bullseye+builder652559327_amd64.changes
./job-build-debian-amd64/output/opensource/debian-bullseye-amd64/mender-artifact_3.10.0~git20220926.4a5b4a0-1+debian+bullseye+builder652559327_amd64.buildinfo
./job-build-debian-amd64/output/opensource/debian-bullseye-amd64/mender-artifact-deb-version
./job-build-debian-amd64/output/opensource/debian-bullseye-amd64/mender-artifact_3.10.0~git20220926.4a5b4a0-1+debian+bullseye+builder652559327_amd64.deb
```

There `/output/opensource/orig/mender-artifact_3.10.0~git20220926.4a5b4a0.orig.tar.gz` is duplicated (and gets overridden in the publish job`) but they should just not be propagated.

I think this can be achieved by selectively `mv` and not `cp -v ... *` [here](https://github.com/mendersoftware/mender-dist-packages/pull/281/files#diff-a696dd766c3d4eb6a58ba77a8c75d4f3888cf19cb5d7332f12a6214f07f0d5d1R124). The script knows which package is building, so it should be able to `mv` only that one and leave the rest for successive runs.

mender-deb-package
```diff
@@ -246,11 +246,14 @@ build_packages() {
 }
 
 copy_packages_to_output() {
+  local binary_only_filter=""
+
+  [[ $DEB_BUILD_TYPE =~ source ]] || binary_only_filter=" -and -not -name *.orig.* "
```
lluiscampos, line 251: But looking at the jobs artifacts I think the filter is not working. In any case I believe we should not filter this way but rather avoid copying them in the first place - see comment above.

lluiscampos (changes requested):

.gitlab-ci.yml
merlin-northern: I agree. how about now? (I took slightly different path)

docker-build-package
```diff
@@ -116,7 +116,12 @@ elif [ "${SAVE_ORIG}" == "true" ]; then
 fi
 mkdir -p "${output_dir}"
 
-[[ "${SAVE_ORIG}" != "true" ]] && cp -v output/opensource/orig/*orig.tar.* "${output_dir}/"
+# we build source packages only for amd64 and for non-commercial
+# (non source closed) components. there is no need to copy the orig
+# file if we do not need it, since later, during publication to the apt-repo
+# we clean them from incoming based on changes file, which refers to orig files
+# only for source builds. See MEN-5879.
+[[ "${SAVE_ORIG}" != "true" && "$commercial" != "true" && "$ARCH" == "amd64" ]] && cp -v output/opensource/orig/*orig.tar.* "${output_dir}/"
```
merlin-northern, line 124: @lluiscampos  this one does (I think) exactly what you said, it gets orig files only when we build source packages.

lluiscampos: Not _exactly_ , because it still copies all orig files (say, mender-artifact, mender-connect, etc) for every single build. My approach was more strict by being explicit on taking only the required orig file. And, on top of it, I recommended using `mv` and not `cp` to have a clean build and avoid the `exclude:` key that we have now in the pipeline job.

Does it make sense?

merlin-northern: but the exclude is so nice, it is obvious that the build package job should not publish the orig directory. and that was your comment:
> build:packages:<distro>:<version>:amd64 produces /output/opensource/<distro>-<version>-amd64/*.orig.tar.gz but does not propagate output/opensource/orig/*.orig.tar.gz.

lluiscampos: > the exclude is so nice,

Sorry, but to me looks like a workaround, not a "nice" design.

> it is obvious that the build package job should not publish the orig directory. and that was your comment: 
> > build:packages:::amd64 produces /output/opensource/--amd64/_.orig.tar.gz but does not propagate output/opensource/orig/_.orig.tar.gz.

It was a comment from high level. I didn't mean achieving that by excluding them (in CI), but designing it so that the files are not passed along (in the script).

.gitlab-ci.yml
```diff
@@ -231,6 +231,8 @@ build:orig:debian:buster:amd64:
   artifacts:
     paths:
       - output/*
+    exclude:
+      - output/opensource/orig/*
```
lluiscampos, line 235: We need to remove the `exclude` in order to verify that the script now manipulates correctly the files (so for example no orig file is left over in `output/opensource/orig/`.

lluiscampos (changes requested):

docker-build-package
merlin-northern: I removed the exclude. but I strongly disagree: exclude is something that is built for this reason.

merlin-northern: @lluiscampos it does not work without the exclude, isnt gitlab caring over them automatically and the exclude is needed?

merlin-northern: effect: all no source builds have orig files carried over. I can explicitly remove them, but come on, this is what we have exclude for.

lluiscampos (approved):
Looks good now :+1: Thank you very much for your patience and the good discussion.

docker-build-package
lluiscampos: Discussed in a call :telephone_receiver:

merlin-northern:
> Looks good now 👍 Thank you very much for your patience and the good discussion.

thank _you_ for your care and great patience with me and my lack of understanding on how gitlab operates.

merlin-northern:
I checked again:
* all required orig files exist
* all changes files where we want source packages refer to orig files
* all referenced orig files exist
* all changes files referring to orig files have a correct check sum
* all checksums of all orig files are consistent (among others: there are no two orig files of the same name having a different check sum)
* every orig file is referenced in at least one changes file (there are no  "rogue" orig files)
* orig directory is not propagated

I cleaned the incoming from both places from the old orig files.

PR #282, opened 2022-09-28
MEN-5941: fix(install-mender.sh): correctly install "latest" for commercial pkgs

lluiscampos:
Previously we were selecting "latest lts" Mender release as the source to decide which version of the commercial packages to install instead of "latest" Mender release, as it used to be before 6f6bc20b.

### DIFF
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -244,23 +244,23 @@ EOF
 }
 
 #
-# Returns the latest Mender LTS release version
+# Returns the latest Mender release version
 # from `docs.mender.io/`
 #
 # Note, requires jq and curl
-function get_latest_mender_lts_release() {
-    [[ $# -ne 0 ]] && { echo >&2 "get_latest_mender_lts_release takes no arguments"; exit 1; }
+function get_latest_mender_release() {
+    [[ $# -ne 0 ]] && { echo >&2 "get_latest_mender_release takes no arguments"; exit 1; }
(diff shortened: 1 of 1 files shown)


scripts/install-mender.sh
```diff
@@ -244,23 +244,23 @@ EOF
...
 #
 # Note, requires jq and curl
-function get_latest_mender_lts_release() {
-    [[ $# -ne 0 ]] && { echo >&2 "get_latest_mender_lts_release takes no arguments"; exit 1; }
+function get_latest_mender_release() {
+    [[ $# -ne 0 ]] && { echo >&2 "get_latest_mender_release takes no arguments"; exit 1; }
     local -r versions_json="$(curl --fail https://docs.mender.io/releases/versions.json)"
-    local -r lts="$(echo "${versions_json}" | jq '.lts[0]')"
-    local -r latest_lts="$(echo "${versions_json}" | jq --raw-output "$(cat <<EOF
+    local -r release_series="$(echo "${versions_json}" | jq '.releases | keys | sort_by( values | split(".") | map(tonumber) )' | jq -r last)"
```
oleorhagen, line 254:
```suggestion
    local -r release_series="$(echo "${versions_json}" | jq '.releases | keys | max)"
```

oleorhagen (approved):
lgtm

scripts/install-mender.sh
lluiscampos: :raised_hands:

PR #283, opened 2022-09-28
Update production branch after MEN-5941

lluiscampos:


lluiscampos:
I tested it in get.mender.io/staging and works as expected :+1:

PR #296, opened 2022-11-15
MEN-6036: fix(mender-client): Remove Update Module generator scritps

lluiscampos:
From next release on, `mender-client` won't install these scripts which
were not meant to be installed on a device in the first place.


lluiscampos:
Best reviewed detecting the copied files with `git diff 135978e0..41804e7a -C -C` :+1:

recipes/mender-client/debian-master/rules
```diff
@@ -7,8 +7,9 @@ override_dh_auto_build:
 	make CGO_CFLAGS="${CGO_CFLAGS}" CGO_LDFLAGS="${CGO_LDFLAGS}" CC="${CC}" \
 	GOOS=linux CGO_ENABLED=1 build
 
-# All install targets except install-systemd
-install_targets = $(shell egrep ^install\-.*: Makefile | grep -v install-systemd | cut -d: -f1)
+install_targets_exclude=install-systemd
+install_targets_all=$(shell grep -E ^install\-.*: Makefile | cut -d: -f1)
```
oleorhagen, line 11: Q: Should not the wildcard be quoted or escaped? :thinking:

recipes/mender-client/debian-3.0.x/mender-client.dirs
```diff
@@ -0,0 +1,5 @@
+/etc/mender
```
oleorhagen, line 1: Q: Are really all of these needed?

recipes/mender-client/debian-3.0.x/postinst
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+set -e
+
+# If the file exists, we understand that meder setup was already run or
```
oleorhagen, line 5: Nit: `meder`

recipes/mender-client/debian-3.0.x/postinst
```diff
@@ -0,0 +1,30 @@
...
+            --retry-poll 300
+
+    # Run the interactive setup wizard
+    else
+        mender setup
+    fi
+
+fi
+
+#DEBHELPER#
```
oleorhagen, line 30: Q: What's this about?

oleorhagen (approved):

recipes/mender-client/debian-master/rules
```diff
@@ -7,7 +7,7 @@ override_dh_auto_build:
 	make CGO_CFLAGS="${CGO_CFLAGS}" CGO_LDFLAGS="${CGO_LDFLAGS}" CC="${CC}" \
 	GOOS=linux CGO_ENABLED=1 build
 
-install_targets_exclude=install-systemd
+install_targets_exclude=install-systemd install-modules-gen
```
kacf, line 10: Just to be clear: We are not now removing the generator scripts from a package which we expect users to install on their build machine, right? Like `mender-client.deb` on your x86_64 desktop machine? From the documentation it looks like we expect people to download the scripts manually, but I'm just making sure.

kacf (approved):

recipes/mender-client/debian-master/rules
lluiscampos: Pre-existing, but I will fix :+1:

recipes/mender-client/debian-3.0.x/postinst
lluiscampos: This is for the debian magic tools to append their code. In our case for example the tool append the systemd service installation and enabling. See the [man page](https://man7.org/linux/man-pages/man7/debhelper.7.html).

recipes/mender-client/debian-3.0.x/mender-client.dirs
lluiscampos: Probably not. Let's do a full review and clean-up in [QA-330](https://tracker.mender.io/browse/QA-330) :+1:

recipes/mender-client/debian-master/rules
lluiscampos: We don't recommend, document nor expect users to have installed `mender-client` on a workstation (iow not device that needs updates). Someone could have done this and we will break his/her flow but I think it is unlikely.

A different topic is if we should package the scripts somewhere else (`mender-artifact-generators`?) and encourage users to install them on their machines. Basically: `apt-get install mender-artifact mender-artifact-generators` and good to go.

In any case, not in this PR :sweat_smile:

PR #302, opened 2023-01-16
feat: Identity and inventory scripts are now conf-files and editable.

ghost:
Changelog: Title
Ticket: MEN-6060


oleorhagen (approved):
Makes sense

PR #312, opened 2023-03-02
QA-530: ci: Use pre-built Docker images

lluiscampos:


lluiscampos:
Depends on:
* https://github.com/mendersoftware/mender-test-containers/pull/302

docker-build-package
```diff
@@ -143,9 +143,11 @@ echo
 docker run --rm \
         --volume $(pwd)/recipes:/recipes \
         --volume $(pwd)/${output_dir}:/output \
+        --volume $(pwd)/mender-deb-package:/script \
         -e MENDER_PRIVATE_REPO_ACCESS_TOKEN="${MENDER_PRIVATE_REPO_ACCESS_TOKEN}" \
         -e MENDER_PRIVATE_REPO_ACCESS_USER="${MENDER_PRIVATE_REPO_ACCESS_USER}" \
         ${IMAGE_NAME_PREFIX}-${DISTRO}-${RELEASE}-${ARCH} \
+        /script \
```
merlin-northern, line 150: I would change that name to something else, as much as I love scripts :) I also love other cool things.

docker-build-package
```diff
@@ -16,7 +16,7 @@
 
 set -e
 
-IMAGE_NAME_PREFIX=mendersoftware/mender-dist-packages:builder
+IMAGE_NAME_PREFIX=mendersoftware/mender-test-containers:mender-dist-packages-builder
```
merlin-northern, line 19: I understand that there is a reason that we are using mender-test-containers with a tag, and not a dedicated image?

merlin-northern (approved):
looks good, left some comments. one general comment: could we have a global check, before the build phase, that `${IMAGE_NAME_PREFIX}-${DISTRO}-${RELEASE}-${ARCH}` image actually exists?

docker-build-package
lluiscampos: Like what? Please suggest something specific.

docker-build-package
lluiscampos: The reason is to be consistent with every other image built in `mender-test-containers` repository, as listed [here](https://github.com/mendersoftware/mender-test-containers/#mender-test-containers).

docker-build-package
merlin-northern: > Like what? Please suggest something specific.

you know naming. let's do it like this: if you are ok with this name, lets keep it.

lluiscampos:
This is the counterpart (and some clean-ups) of https://github.com/mendersoftware/mender-test-containers/pull/302

@mzedel @oldgiova @merlin-northern Could you guys take a look now?

oldgiova (approved):

PR #333, opened 2023-06-26
mender-client-cpp recipes

oleorhagen:


mender-deb-package
```diff
@@ -196,44 +200,76 @@ build_packages() {
...
+    # We want the patches applied for Debian Buster
+    rm -f debian/source/local-options
+    skip_patches_flag=
+  else
+    # Skip applying the patches
+    rm -f debian/patches/series
+  fi
+
+  # Install the builder architecture independent dependencies from the debian/control file
+  mk-build-deps --install --build-indep --tool='apt-get -o Debug::pkgProblemResolver=yes --no-install-recommends --yes' debian/control
```
oleorhagen, line 224: We should probably discuss what we want to do with this?

mender-deb-package
```diff
@@ -196,44 +200,76 @@ build_packages() {
...
-                                -I${ARMV6_TOOLCHAIN_ROOT}/usr/include/arm-linux-gnueabihf \
-                                -pthread \
-                                -I${ARMV6_TOOLCHAIN_ROOT}/usr/include/glib-2.0 \
-                                -I${ARMV6_TOOLCHAIN_ROOT}/usr/lib/arm-linux-gnueabihf/glib-2.0/include/" \
-                    CGO_LDFLAGS="-L${ARMV6_TOOLCHAIN_ROOT}/usr/lib/arm-linux-gnueabihf \
-                                -lssl -lcrypto \
-                                -lz -lgmodule-2.0 -lffi -lmount -lselinux -lpcre2-8 -lblkid -luuid \
-                                -lglib-2.0 -lpcre -lresolv -lgio-2.0 -lgobject-2.0 -lgmodule-2.0" \
+                    CXX=${CROSS_COMPILE}-g++ \
+                    CXXFLAGS=" -Wno-psabi" \
```
oleorhagen, line 254: https://gcc.gnu.org/gcc-7/changes.html

recipes/mender-client/debian-master/control
```diff
@@ -1,24 +1,41 @@
...
-Maintainer: Mender Team <mender@northern.tech>
+Maintainer: The Mender Team <mender@northern.tech>
 Section: admin
 Priority: optional
-Standards-Version: 3.9.2
-Build-Depends: debhelper (>= 9)
+Standards-Version: 4.0.0
+Depends: mender-update, mender-auth
+Build-Depends-Indep: cmake, debhelper (>= 10)
+Build-Depends: liblmdb-dev (>= 0.9.7), libc6-dev (>= 2.14), libboost-dev (>= 1.67.0), libboost-filesystem-dev (>= 1.67.0), libboost-log-dev (>= 1.67.0), libboost-regex-dev (>= 1.67.0), libboost-chrono-dev (>= 1.67.0), libboost-thread-dev (>= 1.67.0), libboost-atomic-dev (>= 1.67.0),  libssl-dev (>= 1.1.1), libarchive-dev (>= 3.2.2)
```
oleorhagen, line 8: This was added mostly due to `mk-build-deps`, but we can decide what we want to do with this.

recipes/mender-client/debian-master/rules
```diff
@@ -1,42 +1,24 @@
...
 
 override_dh_auto_test:
 	true
 
-override_dh_strip:
-	${CROSS_COMPILE}-strip --remove-section=.comment --remove-section=.note \
-	debian/mender-client/usr/bin/mender
+# This is required as we cheat when we install the debian packages for raspbian
+override_dh_shlibdeps:
+	dh_shlibdeps --dpkg-shlibdeps-params=--ignore-missing-info
```
oleorhagen, line 24: This, or we can install all the files from the debian packages, when we install them manually

recipes/mender-client/debian-master/source/local-options
```diff
@@ -0,0 +1 @@
+skip-patches
```
oleorhagen, line 1: In general, we don't want to apply the boost beast patches

docker-build-package
```diff
@@ -18,9 +18,10 @@ set -e
 
 IMAGE_NAME_PREFIX=registry.gitlab.com/northern.tech/mender/mender-test-containers:mender-dist-packages-builder
 
+# TODO (MEN-6395) - hardcoded to oleorhagen whilst this is a feature branch
```
kacf, line 21: This need to refer to a future task. I created https://northerntech.atlassian.net/browse/MEN-6569.

install-armv6-cross-toolchain
```diff
@@ -0,0 +1,163 @@
+#! /bin/bash
+
+set -ex
+
+# To provide support for Raspberry Pi Zero W a toolchain tuned for ARMv6 architecture must be used.
+# https://northerntech.atlassian.net/browse/MEN-2399
+# ARMV6_TOOLCHAIN_ROOT="/armv6-eabihf--glibc--stable-2020.08-1"
+TOOLCHAIN_PREFIX=/opt/
+ARMV6_TOOLCHAIN_ROOT="${ARMV6_TOOLCHAIN_ROOT:?No toolchain set}"
+RASPBIAN_VERSION="${VERSION:?No Raspbian version set}"
```
kacf, line 10: Using spaces here makes me nervous. I also think it might hide the message you're trying to convey here ("No" not found). Underscores?

Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+    fi
+
+# TODO: Figure out why these packages are required
+RUN if [ "${DISTRO}" = "debian" -a "${ARCH}" = "armhf" ]; then \
+        apt-get install -y libmount1:armhf zlib1g:armhf; \
+    fi
+
+# Prepare the deb-package script
+COPY mender-deb-package /usr/local/bin/
+ENTRYPOINT  ["/usr/local/bin/mender-deb-package"]
```
lluiscampos, line 80: The `Dockerfile`, together with the new script `install-armv6-cross-toolchain`, needs to be submitted to `mender-test-containers` instead ([here](https://github.com/mendersoftware/mender-test-containers/tree/master/mender-dist-packages-building)).

lluiscampos (changes requested):

kacf (commented):
I started reviewing, but aborted once I realized that this should be consolidated with the move of the Dockerfile to mender-test-containers. There are some comments, but I will resume the review once the consolidation is done, since it will be much easier.

install-armv6-cross-toolchain
oleorhagen: Hmm, I think you misunderstand, this is quite common usage:

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

kacf: Ah, question mark! You're right, I misread that. (And a much better approach)

mender-deb-package
```diff
@@ -135,6 +135,9 @@ prepare_recipe() {
     if [ "${debian_recipe}" = "debian-2.1.x" ]; then
       # Skip for 2.1.x, special handling in debian/rules
       true
+    elif [ "${debian_recipe}" = "debian-master" ]; then
+      # skip
+      true
```
kacf, line 140: Shouldn't this have a task to remove? It's ok that we don't need it now, since we have no daemon mode, but we shouldn't leave it like this.

mender-deb-package
kacf: This is way above my head (I hope Lluis understands this! 🙂)

mender-deb-package
```diff
@@ -196,44 +199,73 @@ build_packages() {
...
-                    ${sign_flags} \
-                    --build=$DEB_BUILD_TYPE
+        CC="$CROSS_COMPILE-gcc" \
+        PKG_CONFIG_PATH="/usr/lib/arm-linux-gnueabihf/pkgconfig/" \
+        GOARCH=arm \
+        dpkg-buildpackage -aarmhf \
+        ${sign_flags} \
+        "${dpkg_build_ignore_flag}" \
+        ${skip_patches_flag} \
+        --build=$DEB_BUILD_TYPE
```
kacf, line 268: Nitpick: Indentation.

recipes/mender-client/debian-master/control
kacf: It's ok, IMHO.

recipes/mender-client/debian-master/control
```diff
@@ -1,19 +1,34 @@
...
+Depends: ${shlibs:Depends}, ${misc:Depends}
+Description: Mender update
+ Mender is an open source over-the-air (OTA) software updater for embedded Linux devices.
+
+Package: mender-auth
+Conflicts: mender
+Architecture: any
+Depends: ${shlibs:Depends}, ${misc:Depends}
+Description: Mender auth
+ Mender is an open source over-the-air (OTA) software updater for embedded Linux devices.
```
kacf, line 30: "Mender-auth enables centralized authentication against a Mender server".

recipes/mender-client/debian-master/control
```diff
@@ -1,19 +1,34 @@
...
+Depends: mender-client, mender-update
 Description: Mender client
  Mender is an open source over-the-air (OTA) software updater for embedded Linux devices.
 
+Package: mender-update
+Conflicts: mender
+Architecture: any
+Depends: ${shlibs:Depends}, ${misc:Depends}
+Description: Mender update
+ Mender is an open source over-the-air (OTA) software updater for embedded Linux devices.
```
kacf, line 23: "Mender-update"

recipes/mender-client/debian-master/mender-auth.install
```diff
@@ -0,0 +1,2 @@
+/lib/systemd/system/mender-authd.service
+/usr/bin/mender-auth
```
kacf, line 2: I think `/usr/share/mender/identity` should be here as well.

recipes/mender-client/debian-master/patches/boost-1.6.x-beast.patch
```diff
@@ -0,0 +1,132 @@
...
+-#include <boost/beast/core/stream_traits.hpp>
++#include <boost/beast/core/type_traits.hpp>
+ #include <boost/asio.hpp>
+ #include <common/http.hpp>
+ 
+@@ -94,7 +94,6 @@ Client::Client(ClientConfig &client, eve
+ 	resolver_(GetAsioIoContext(event_loop)),
+ 	stream_(GetAsioIoContext(event_loop), client.ctx_),
+ 	body_buffer_(HTTP_BEAST_BUFFER_SIZE) {
+-	response_buffer_.reserve(body_buffer_.size());
```
kacf, line 54: Have you tested this? I suspect this will generate errors of the `need_buffer` kind.

recipes/mender-client/debian-master/patches/boost-1.6.x-beast.patch
```diff
@@ -0,0 +1,132 @@
...
+ 	// Don't enforce limits. Since we stream everything, limits don't generally apply, and
+ 	// if they do, they should be handled higher up in the application logic.
+@@ -188,8 +187,8 @@ void Client::ResolveHandler(
+ 
+ 	weak_ptr<Client> weak_client(stream_active_);
+ 
+-	beast::get_lowest_layer(this->stream_)
+-		.async_connect(
++	boost::asio::async_connect(
++			this->stream_.next_layer(),
```
kacf, line 65: Both of these approaches are available in later versions too. We could switch the main code to this. It's not really that bad.

recipes/mender-client/debian-master/rules
```diff
@@ -1,42 +1,24 @@
...
+#
+#DH_VERBOSE=1
+# see FEATURE AREAS in dpkg-buildflags(1)
+#export DEB_BUILD_MAINT_OPTIONS = hardening=+all
+
+# see ENVIRONMENT in dpkg-buildflags(1)
+# package maintainers to append CFLAGS
+#export DEB_CFLAGS_MAINT_APPEND  = -Wall -pedantic
+# package maintainers to append LDFLAGS
+#export DEB_LDFLAGS_MAINT_APPEND = -Wl,--as-needed
```
kacf, line 11: Remove.

recipes/mender-client/debian-master/rules
```diff
@@ -1,42 +1,24 @@
...
-
-	# Install scripts/version file
-	install -d -m 755 debian/mender-client/etc/mender/scripts/
-	echo -n "3" > debian/mender-client/etc/mender/scripts/version
-
-	# MEN-6060: Make identity and inventory scripts into conf files so that
-	# they can be modified without risking that they get lost on an
-	# upgrade. To follow Debian policy, we must then put them in /etc and
-	# symlink to them, we can not simply mark them as conffiles:
-	# https://www.debian.org/doc/debian-policy/ch-files.html#location
```
kacf, line 24: Shouldn't this be kept?

tests/test_package_client_cpp.py
```diff
@@ -0,0 +1,288 @@
...
+    },
+    #
+    # device identity files
+    #
+    {
+        "name": "/usr/share/mender/inventory",
+        "type": "directory",
+    },
+    {
+        "name": "/usr/share/mender/inventory/mender-device-identity",
```
kacf, line 110: This is misspelled: Identity in an inventory folder.

tests/test_package_client_cpp.py
```diff
@@ -0,0 +1,288 @@
...
+        verify_file_exists(ssh_connection, all_files)
+        # Northern.tech copyright file
+        ssh_connection.run("test -f /usr/share/doc/mender-client/copyright")
+        result = ssh_connection.run(
+            "tail -n +2 /usr/share/doc/mender-client/copyright | md5sum"
+        )
+        assert result.stdout.split(" ")[0] == self.expected_copyright_from_l2_md5sum
+
+@pytest.mark.cppclient
+class TestPackageMenderClientDefaults(PackageMenderClientChecker):
```
kacf, line 192: Nice tests!

mender-deb-package
```diff
@@ -196,44 +199,73 @@ build_packages() {
...
         ${sign_flags} \
+        "${dpkg_build_ignore_flag}" \
         --build=$DEB_BUILD_TYPE
       ;;
 
     armhf)
       if [ "$OS_DISTRO" = "debian" ]; then
         # Custom toolchain to support ARMv6
+        PATH="${ARMV6_TOOLCHAIN_ROOT}/bin:$PATH" \
         CROSS_COMPILE="arm-buildroot-linux-gnueabihf" \
```
kacf, line 247: Nitpick: Indentation.

kacf (commented):
I have to admit that I only barely follow what is happening in the package building. I think you two are both bigger experts than me. I pointed out what I could though.

Thanks for sticking with it!

recipes/mender-client/debian-master/rules
oleorhagen: Mby, dunno

oleorhagen:
Goes along with:
https://github.com/mendersoftware/mender-test-containers/pull/361

oleorhagen:
Alright, @lluiscampos @kacf  I _finally_ got it to work after quite a brawl! :tada: 

I am pretty tired now, it's been a long day. I don't have the energy to go over and polish this for review right now, but the general direction should  be there, so try and have a look still.

If I show up to work tomorrow, and I hear something along the lines of: `this file is not in the right place, I can't review` we're gonna have a brawl. Elon Musk and Zuckerberg style.

Good morning, and see you eventually!

oleorhagen:
Ps: I don't expect the tests to pass as of now, but so be it.

mender-deb-package
```diff
@@ -55,9 +55,9 @@ checkout_repo_clean_local_path() {
 
 checkout_repo() {
   local -r repo_path=$(checkout_repo_clean_local_path "${REPO_URL}")
-  git clone ${REPO_URL} ${repo_path}
+  git clone --depth=1 --branch ${VERSION} ${REPO_URL} ${repo_path}
```
lluiscampos, line 58: Careful, this won't work with tags :boom: 

Either leave it as it was (full clone, then checkout) or use the mender-qa way (inspiration [here](https://github.com/mendersoftware/mender-qa/blob/f9dbfd797299aff75da060128012edd667d2c74e/gitlab-pipeline/stage/init.yml#L54))

mender-deb-package
lluiscampos: We should skip this on debian/armhf build, where we have tracked down all the dependencies (in [here](https://github.com/mendersoftware/mender-test-containers/pull/361/files#diff-8843be8793b8fa8003b51d477dfa6943c2ea7ac81a8cddb4e327c59f98dbd07dR14)).

For the rest we can leave it (which is a better approach than having the dependencies installed in the Dockerfile before hand)

mender-deb-package
```diff
@@ -196,56 +199,90 @@ build_packages() {
...
+  fi
+
+  # Install the builder architecture independent dependencies from the debian/control file
+  mk-build-deps --install --build-indep --tool='apt-get -o Debug::pkgProblemResolver=yes --no-install-recommends --yes' debian/control
+
+  # HOST - the machine arch we are building _for_
+  HOST_ARCH=${ARCH}
+
+  dpkg --add-architecture ${ARCH}
+  apt update
```
lluiscampos, line 229: This we don't need. It is [already there](https://github.com/mendersoftware/mender-test-containers/pull/361/files#diff-4fe9a1f0c62ee8dc7cc0d71d9560b2283ce14c5f4b9988405b60a0dd3c37b0cfL25) for the relevant builds (all except debian/armhf)

mender-deb-package
```diff
@@ -196,56 +199,90 @@ build_packages() {
...
+
+  mk-build-deps --install \
+                --build-dep \
+                --tool='apt-get -o Debug::pkgProblemResolver=yes --no-install-recommends --yes' \
+                --host-arch ${HOST_ARCH} \
+                debian/control
```
lluiscampos, line 235: Similarly, this should not be needed for `debian/armhf`

mender-deb-package
lluiscampos: I would add this link as a comment in the script, somewhere.

recipes/mender-client/debian-master/control
```diff
@@ -1,24 +1,39 @@
...
-# debian bullseye Depends: libc6 (>= 2.14), libglib2.0-0 (>= 2.26.0), liblzma5 (>= 5.1.1alpha+20120614), libssl1.1 (>= 1.1.1)
-# debian buster Depends: libc6 (>= 2.14), libglib2.0-0 (>= 2.26.0), liblzma5 (>= 5.1.1alpha+20120614), libssl1.1 (>= 1.1.1)
+Depends: mender-client, mender-update
 Description: Mender client
  Mender is an open source over-the-air (OTA) software updater for embedded Linux devices.
 
+Package: mender-update
+Conflicts: mender
+Architecture: any
+Depends: ${shlibs:Depends}, ${misc:Depends}
```
lluiscampos, line 21: `mender-update` should recommend (or depend on?) `mender-auth` :thinking:

recipes/mender-client/debian-master/rules
lluiscampos: Yes, we should keep it. We can either keep this hook here or modify the install targets in our CMake to install the files in `etc` and create the symlinks from `/usr/share`

recipes/mender-client/debian-master/rules
lluiscampos: This is incorrect but I don't know how to solve it.

So previously we had for debian/armhf harcoded the dependencies ([here](https://github.com/mendersoftware/mender-dist-packages/pull/333/files#diff-2ae0f2ab603d72f1d5c372cd757066bbfb5399baa282f80e85172ebcc6abeabaL11-L13)) while for the rest we used the  `shlibs:Depends` magic. Now however we have set up `shlibs:Depends` for all and then changing here the target to make it "work" on debian/armhf.

On the high level, we should only make exceptions for `debian/armhf` and leave the other cases to build with the `shlibs:Depends`, but I don't have a good suggestion to make this work in the context of this PR.

tests/test_package_client_cpp.py
lluiscampos: Indeed

lluiscampos (changes requested):
There are few things we should clean up IMO. I know that these areas are brittle slow to test, so feel free to come up with counter arguments :)

Generally speaking looks good. Brave work :muscle:

recipes/mender-client/debian-master/patches/boost-1.6.x-beast.patch
oleorhagen: Keep in mind that they are both available, but are not really the same. Only in the case we have one layer wrapped.

kacf: Yes, but we expect the next layer to be tcp. In fact, there is a different function one can use to get the lowest layer in 1.67, but this didn't work, which could mean that there is another, perhaps internal, layer after tcp. So I'd argue that `next_layer` is what we should use.

recipes/mender-client/debian-master/rules
oleorhagen: It's only ignoring the info files, not all dependencies, so it's not super critical. I suspect it's because we only install the library files themselves in our unpackaging, but don't have the time to find out right now.

For now i think this is Ok though

recipes/mender-client/debian-master/rules
oleorhagen: Leaving this for now

recipes/mender-client/debian-master/patches/boost-1.6.x-beast.patch
oleorhagen: Not against it. Just good to keep in mind for the future

oleorhagen:
Alright, @kacf @lluiscampos new edition.

This is now building for bullseye and buster, and manually tested on buster.

It is still not completely done, but it is at the point where we can use it for publishing tomorrow :tada: 

I have commented out the files in the `check_files` test that are currently not installed.

See you in the morning

recipes/mender-client/debian-master/rules
oleorhagen: fixed

mender-deb-package
oleorhagen: I'm guessing the submodules can be kept at `depth=1` though

oleorhagen: Hmm, the docs says it should work just fine with tags as well :thinking: 

```
--branch <name>

    Instead of pointing the newly created HEAD to the branch pointed to by the cloned repository’s HEAD, point to <name> branch instead. In a non-bare repository, this is the branch that will be checked out. --branch can also take tags and detaches the HEAD at that commit in the resulting repository.
```

Am I missing something obvious?

lluiscampos (approved):
Few pending things but are not blockers IMO as we can investigate and follow-up before final tags.

Impressive tenacity in this work, Ole :medal_military:

mender-deb-package
oleorhagen: Resolved, this works

recipes/mender-client/debian-master/rules
oleorhagen: Fixed locally by installing all the files during container build

recipes/mender-client/debian-master/control
oleorhagen: I'm still a little undecided tbh...

Let's kick this down the road for now :sweat_smile:

PR #337, opened 2023-06-28
ci: Pin the builder image to an specific version

lluiscampos:
While introducing support for C++ sources we have broken support for golang sources. See MEN-6579.

While we fix the mentioned issue, we can use a fixed version on master to keep the pipelines running.


lluiscampos:
Note to self: we need to retag all images with `-pin` suffix and amend this PR.

oldgiova (approved):

kacf (approved):

PR #338, opened 2023-06-28
ci: Skip building source packages

lluiscampos:
This will also have to be re-enabled with MEN-6579.

### DIFF
--- a/docker-build-package
+++ b/docker-build-package
@@ -129,10 +129,10 @@ else
     BUILD_TYPE=any
 fi
 
-if [ "$commercial" != "true" -a "$ARCH" = "amd64" ]; then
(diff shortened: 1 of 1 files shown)


lluiscampos:
@kacf This should workaround the problem that we saw earlier when building the first alpha

lluiscampos:
Pipeline building the alpha tag @ [Pipeline 915096276](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/915096276)

kacf (approved):
Should we create a ticket to fix this properly?

lluiscampos:
> Should we create a ticket to fix this properly?

I extended the scope of MEN-6579 to cover also this workaround :+1:

PR #339, opened 2023-06-29
ci: Include `.ddeb` in the publish job

lluiscampos:


.gitlab-ci.yml
```diff
@@ -225,7 +225,7 @@ test:acceptance:cpp:
   script:
     # Upload files: first .buildinfo and .deb, then .changes
     - for d in output/opensource/*; do
-    -   find "$d" -name '*.buildinfo' -o -name '*.deb' -o -name '*.orig.tar.*' -o -name '*.debian.tar.*' -o -name '*.dsc' | while read -r file; do
+    -   find "$d" -name '*.buildinfo' -o -name '*.deb' -o -name '*.orig.tar.*' -o -name '*.debian.tar.*' -o -name '*.dsc' -o -name '*.ddeb' | while read -r file; do
```
merlin-northern, line 228: I did not know we were building the debug debs? since when is that happening?

merlin-northern (approved):

.gitlab-ci.yml
lluiscampos: Note that this PR is going into the feature branch.

With the new C++ client we have built this packages. We might decide to remove these later, but in any case taking them into account here won't harm.

PR #349, opened 2023-08-31
Men 6579 re enable golang builds in cpp branch

oleorhagen:


mender-deb-package
```diff
@@ -246,32 +248,38 @@ EOF
...
-              CXX=${CROSS_COMPILE}-g++ \
-              GOARCH=arm \
-              GOARM=6 \
-              dpkg-buildpackage -aarmhf \
-              ${sign_flags} \
-              "${dpkg_build_ignore_flag}" \
-              ${skip_patches_flag} \
+            # For some reason CMake find_package does not like it when this dir is missing
+            SYSROOT="/opt/cross-pi-gcc-10.3.0-0/arm-linux-gnueabihf/libc/"
+            mkdir /opt/cross-pi-gcc-10.3.0-0/arm-linux-gnueabihf/libc/include
```
lluiscampos, line 270: Why `bullseye` needs an explicit `mkdir` here while `buster` does not? Could we just do it in both places (with `mkdir -p`) for consistency?

lluiscampos: Suggested beautification:
```suggestion
            mkdir "${SYSROOT}/include"
```

lluiscampos (approved):
Looks good!

Thank you for the style alignment :sparkles:

PR #353, opened 2023-09-14
feat: mender-app-update-module packages

merlin-northern:
Changelog: Title
Ticket: MEN-6196


docker-build-package
```diff
@@ -106,7 +114,7 @@ if [ "$arch_indep" = "true" -a "$ARCH" != "amd64" ]; then
     exit 0
 fi
 
-echo "Building $recipe_name..."
+echo "Building $recipe_name... $*"
```
lluiscampos, line 117: Thank you. Good idea.

I suggest thought a separate commit for it and something more explicit like
```suggestion
echo "Building $recipe_name with arguments: $*"
```

recipes/mender-app-update-module/debian-master/control
```diff
@@ -0,0 +1,15 @@
...
+Maintainer: Mender Team <mender@northern.tech>
+Section: admin
+Priority: optional
+Standards-Version: 3.9.2
+Build-Depends: debhelper (>= 9)
+
+Package: mender-app-update-module
+Architecture: all
+Depends: ${misc:Depends}
+Suggests: mender-client
```
lluiscampos, line 11: I think it should `depend` on the client. Can the update module run as standalone in any scenario?

Also, please add the minimum version. In theory it could work with >2.0 (when we introduced update modules) but I would write the oldest version that we have tested it against.

recipes/mender-app-update-module/debian-master/control
```diff
@@ -0,0 +1,15 @@
...
+Description: Mender Application Update Modules
+ Mender provides application updates via modules. They provide
+ a way to update and manage container-based deployments.
```
lluiscampos, line 14: Nice :sparkles:

recipes/mender-app-update-module/debian-master/rules
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/make -f
+
+%:
+	dh $@
+
+override_dh_auto_build:
+	true
```
lluiscampos, line 7: Why do we need to override this one? It should work (iow do nothing) as long as there is a Makefile

recipes/mender-app-update-module/debian-master/rules
```diff
@@ -0,0 +1,16 @@
...
+
+override_dh_auto_install:
+	DESTDIR=debian/mender-app-update-module make install
+
+override_dh_auto_test:
+	true
+
+override_dh_strip:
+	true
```
lluiscampos, line 16: Also these two should be able to run with defaults, why do you need to override?

lluiscampos (commented):
It looks good. Well done!

Some questions below before I can give the final :heavy_check_mark:

recipes/mender-app-update-module/debian-master/rules
merlin-northern: fixed

recipes/mender-app-update-module/debian-master/rules
merlin-northern: fixed

recipes/mender-app-update-module/debian-master/control
merlin-northern: fixed, and also: this meant updating the test. I did Suggests: on purpose, but if you think it should be Depends: I am ok with it.

docker-build-package
merlin-northern: fixed

lluiscampos (approved):

PR #363, opened 2023-10-05
chore: Migrate from deprecated distutils

lluiscampos:
Clears `DeprecationWarning`. See:
https://peps.python.org/pep-0632/#migration-advice

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,7 +13,7 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
-from distutils.version import LooseVersion
+import packaging.version
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -13,7 +13,7 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
-from distutils.version import LooseVersion
+import packaging.version
```
oleorhagen, line 16: S: from packaging.version import Version, and then change the below, but YMMV

oleorhagen (approved):

tests/conftest.py
lluiscampos: We would need to import also `InvalidVersion`, so I think is is cleaner this way TBH. (Another not too good reason is that I basically copied from [here](https://github.com/mendersoftware/meta-mender/commit/6d2ca0800387a17fccd9b51b9f82f6c17903fd9d)).

PR #375, opened 2023-11-28
MEN-6858: Update `Depends` for supported `mender-connect` recipes

lluiscampos:
It requires either `mender-client` or `mender-auth`, which is one of the new packages being distributed as part of Mender client 4.0 onwards.

### DIFF
--- a/recipes/mender-connect/debian-master/control
+++ b/recipes/mender-connect/debian-master/control
@@ -7,9 +7,9 @@ Build-Depends: debhelper (>= 9)
 
 Package: mender-connect
 Architecture: any
-Depends: mender-client (>3.2.0), ${shlibs:Depends}, ${misc:Depends}
-# debian bullseye Depends: mender-client (>= 3.1.99), libc6 (>= 2.3.2), libglib2.0-0 (>= 2.26.0)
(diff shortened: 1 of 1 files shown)


lluiscampos:
The first commit has no code changes. Best reviewed with: `git show -C -C 90a1d1`

recipes/mender-connect/debian-master/control
```diff
@@ -7,7 +7,7 @@ Build-Depends: debhelper (>= 9)
 
 Package: mender-connect
 Architecture: any
-Depends: mender-client (>3.2.0), ${shlibs:Depends}, ${misc:Depends}
+Depends: mender-client (>4.0.0), ${shlibs:Depends}, ${misc:Depends}
```
kacf, line 10: mender-auth, not mender-client, right?

kacf (changes requested):

recipes/mender-connect/debian-master/control
kacf: From [this page](https://www.debian.org/doc/debian-policy/ch-relationships.html):

> the package names listed may also include lists of alternative package names, separated by vertical bar (pipe) symbols

Sounds exactly like what we need, doesn't it?

PR #381, opened 2023-12-13
fix: Fix the Boost dependency for legacy OS

lluiscampos:
For these, use `MENDER_DOWNLOAD_BOOST=ON` to let `CMake` download a compatible version of Boost.

The intention with implementing it as a patch is to avoid diverging the `rules` for current and legacy builds.


recipes/mender-client/debian-master/rules.legacy.patch
```diff
@@ -0,0 +1,12 @@
+--- rules	2023-12-13 09:06:16.575373793 +0100
++++ rules.legacy	2023-12-13 09:07:10.555779873 +0100
+@@ -6,7 +6,8 @@
+ override_dh_auto_configure:
+ 	dh_auto_configure -- \
+ 		-D CMAKE_BUILD_TYPE=Release \
+-		-D BOOST_ALL_DYN_LINK=ON
++		-D BOOST_ALL_DYN_LINK=ON \
```
kacf, line 8: I suspect this might conflict with the `MENDER_DOWNLOAD_BOOST` flag, since it internally uses static linking (it has to, because there are no system libraries to link to).

kacf (approved):

recipes/mender-client/debian-master/rules.legacy.patch
lluiscampos: Do you mean we should should then replace `BOOST_ALL_DYN_LINK` with `MENDER_DOWNLOAD_BOOST` in the legacy builds?

lluiscampos: CI seems to be happy. If there is a conflict, do you expect it to shown at build time or at run time?

kacf: Yes. Possibly also for the normal builds. I removed it from the cmake files because it conflicted with the `MENDER_DOWNLOAD_BOOST` flag, and have yet to see any problems with it.

mender-deb-package
```diff
@@ -149,6 +149,13 @@ prepare_recipe() {
     cp support/mender-monitor.service debian/
   fi
 
+  # For legacy OS, patch the debian/rules file for mender-client
+  if [ "${DEB_PACKAGE}" = "mender-client" ]; then
```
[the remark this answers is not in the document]
lluiscampos: Yes, it is. How else can we organize it? This is not really patching the source code but the recipe. The alternative is commented out code like we do for `control` in the rest of the packages.

Note that the `skip-patches` is not in use anymore, we don't have patches to apply in any case.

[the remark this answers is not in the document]
oleorhagen: Yeh, I noticed too. I removed it locally actually :see_no_evil: 

Maybe just open a PR. Then we can keep this as is?

recipes/mender-client/debian-master/control
```diff
@@ -20,19 +22,22 @@ Package: mender-update
...
 Architecture: any
 Depends: ${shlibs:Depends}, ${misc:Depends}
+# debian buster Depends: libc6 (>= 2.28), libdbus-1-3 (>= 1.9.14), libgcc-s1 (>= 3.5), libssl1.1 (>= 1.1.1), libstdc++6 (>= 9)
 Description: Mender auth
  Mender-auth enables centralized authentication against a Mender server.
 
 Package: mender-client-dev
 Architecture: all
 Depends: mender-client (= ${binary:Version}), ${misc:Depends}
+# debian buster Depends: mender-client (= ${binary:Version})
```
kacf, line 40: `#` is part of the syntax?? Really weird...

kacf (approved):

recipes/mender-client/debian-master/control
lluiscampos: No, this is our hack to hard-code dependencies and not invoke Debian Helpers. See [here](https://github.com/mendersoftware/mender-dist-packages/blob/f68ac9acf101e0dd9803fe4f7498b41155975315/mender-deb-package#L157).

It is what we have been using for all recipes until the C++ client introduction when we _tried_ to be more streamlined.

kacf: Ah, this is our thing, got it.

PR #382, opened 2023-12-13
Miscellaneous fixes for Mender client C++ package(s)

lluiscampos:


lluiscampos:
The first three commits are from https://github.com/mendersoftware/mender-dist-packages/pull/381 - please review the rest.

tests/test_package_client_cpp.py
```diff
@@ -230,11 +231,6 @@ def test_install_configure_start(
 
         self.check_mender_client_version(setup_tester_ssh_connection, mender_version)
 
-        # Default setup expects ServerURL hosted.mender.io
-        # TODO - No mender-setup yet
-        # result = setup_tester_ssh_connection.sudo("cat /etc/mender/mender.conf")
-        # assert '"ServerURL": "https://hosted.mender.io"' in result.stdout
```
kacf, line 237: Why don't we want to test this?

kacf (approved):

tests/test_package_client_cpp.py
lluiscampos: It is tested for `mender-setup` package. See [here](https://github.com/mendersoftware/mender-dist-packages/blob/3ad8a339919bfbf57d1e420d5db8e8f6fb28fd5f/tests/test_package_setup.py#L60).

kacf: Right, of course!

kacf (approved):

mender-deb-package
```diff
@@ -134,12 +134,13 @@ prepare_recipe() {
     if [ "${debian_recipe}" = "debian-2.1.x" ]; then
       # Skip for 2.1.x, special handling in debian/rules
       true
-    elif [ "${debian_recipe}" = "debian-master" ]; then
-      # skip
-      true
-    else
+    elif [[ "${debian_recipe}" =~ ^debian-2.*  || "${debian_recipe}" =~ ^debian-3.* ]]; then
```
oleorhagen, line 137: Q: Shouldn't the `wildcard` be quoted?

oleorhagen: You can also do: ` =~ ^debian-(2|3).*` to make it shorter :)

recipes/mender-client/debian-master/mender-client-dev.install
```diff
@@ -1 +1 @@
-Documentation/io.mender.*.xml /usr/share/dbus-1/interfaces/
+/usr/share/dbus-1/interfaces/*
```
oleorhagen, line 1: Nice!

recipes/mender-client/debian-master/control
```diff
@@ -3,8 +3,6 @@ Maintainer: The Mender Team <mender@northern.tech>
 Section: admin
 Priority: optional
 Standards-Version: 4.0.0
-Depends: mender-update, mender-auth
-# debian buster Depends: mender-update, mender-auth
 Build-Depends-Indep: cmake, debhelper (>= 10)
 Build-Depends: liblmdb-dev (>= 0.9.7), libc6-dev (>= 2.14), libboost-dev (>= 1.67.0), libboost-filesystem-dev (>= 1.67.0), libboost-log-dev (>= 1.67.0), libboost-regex-dev (>= 1.67.0), libboost-chrono-dev (>= 1.67.0), libboost-thread-dev (>= 1.67.0), libboost-atomic-dev (>= 1.67.0),  libssl-dev (>= 1.1.1), libarchive-dev (>= 3.2.2), libdbus-1-dev
```
oleorhagen, line 7: I guess now, technically, the `boost` versions here are not correct either.

tests/test_package_client_cpp.py
```diff
@@ -155,6 +147,13 @@ def check_installed_files(self, ssh_connection):
         )
         assert result.stdout.split(" ")[0] == expected_copyright_from_l2_md5sum
 
+    def check_systemd_service(self, ssh_connection):
+        result = ssh_connection.run("systemctl is-enabled mender-authd")
```
oleorhagen, line 151: Nice!

recipes/mender-client/debian-master/rules
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/make -f
 
 %:
-	dh $@
+	dh $@ --with systemd
```
oleorhagen, line 4: Hmm, if I understand the documentation right, this should have already been enabled, since we are using `compat level >= 10`. But obviously not, since it wasn't working before? :thinking:

recipes/mender-client/debian-master/control
lluiscampos: You are right. But changing it will required yet another special handling for old Debian and old Ubuntu to ignore these. I think we let them be.

recipes/mender-client/debian-master/rules
lluiscampos: It didn't work without this. The service files were installed but not enabled (iow no symlink from `/etc`)

oleorhagen: Yeh, I believe you (since obviously it wasn't before). Just wanted to point out that it doesn't match with what I got out of the documentation at least

mender-deb-package
lluiscampos: At least on my version of bash, when quoting it does string comparison and not regex. There are also other non-quoted regex here in this file.

I will modify it with your fancy suggestion but not quote it :+1:

oleorhagen: Yup, I guess it makes sense. Bash is `magic` that is for sure :magic:

PR #388, opened 2023-12-18
ci: Fix missing newline escape

lluiscampos:
This mistake came in from the `feature-c++-client` branch but because we didn't had there privileges to run commercial tests it went unnoticed.

Thank you Ole for your over-the-shoulder review.


tranchitella (approved):

lluiscampos:
Merging without waiting for the pipeline - this fix does not run in PRs

PR #392, opened 2023-12-19
ci: Restore functionality to skip tests

lluiscampos:
The `rules` were being overridden in the child jobs and was not possible to skip the tests anymore.


oleorhagen (approved):

oleorhagen:
I thought it merged the fields when overriding, but then appareantly not

PR #394, opened 2023-12-20
fix: Install by default `mender-flash` and `mender-setup`

lluiscampos:
The reasoning is that the packages should behave as they used to:
* `mender-update` should have the read-on-write optimization as it used to (iow, install `mender-flash`) and to be able to interactively set the configuration fields (with `mender-setup`).
* `mender-auth` should also be able to be configured interactively.
* `mender-client` meta package should provide everything that the old client had, including the snapshot feature with `mender-snapthot`.

Ticket: MEN-6920
Changelog: None


oleorhagen (approved):
Makes sense

lluiscampos:
The last update is only an amend on the commit message. The previous pipeline was [here](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/1114123346) :green_circle:

PR #404, opened 2024-01-18
MEN-6965: Support downgrade of `mender-client` from 4 to v3

lluiscampos:
* fix: Add placeholders for debhelper for `mender-client` master
    
    These are needed in order to the systemd services to be started and
    stopped automatically by `apt-get install` or `remove`.

* fix: Support downgrade of `mender-client` from 4 to v3
    
    This commit consist on various parts:
    * Introduce an epoch for all `mender-client` package builds so that we
      can publish a `1:3.5.2` that will be upgradable from already
      published `4.0.0` version. See:
      https://www.debian.org/doc/debian-policy/ch-controlfields.html#version
    * Add `preinst` and `postrm` scripts on the 3.5.x series to detect
      an "upgrade" from v4 and:
        * Take back ownership of the configuration files
        * Disable `mender-authd` and `mender-updated` systemd services as
          this was missing from the first publishing of the packages (see
          previous commit)
    * Mark the `mender-client` 3.5.x series package as conflict, replaces
      and provides of the new packages, so that they get removed when
      installing `mender-client`. See:
      https://www.debian.org/doc/debian-policy/ch-relationships.html#replacing-whole-packages-forcing-their-removal
    * New tests for the different upgrade paths. Note that these make only
      sense when building a 3.5.2 or newer 3.5.x *tag*.

### DIFF
--- a/mender-deb-package
+++ b/mender-deb-package
@@ -87,16 +87,26 @@ get_deb_distribution() {
   fi
 }
 
+# This function declares three variables for the rest of the script to use:
+#   DEB_VERSION          The full Debian technical version
+#   DEB_VERSION_NO_REV   The Git version of the package, either tag or X.Y.Z~gitDATE
+#   DEB_VERSION_NO_EPOCH The Debian version minus the epoch, matching the string in the filename
 get_deb_version() {
   # Create a version from Git.
-  #  - For Git tags: X.Y.Z-<debian_suffix>
-  #  - For master: X.Y.Z~git<commit-date>.<commit-sha>-<debian_suffix>+b<BUILD_ID>
+  #  - For Git tags: <debian-epoch>X.Y.Z-<debian_suffix>
+  #  - For master: <debian-epoch>X.Y.Z~git<commit-date>.<commit-sha>-<debian_suffix>+b<BUILD_ID>
   #     where X.Y.Z is latest tag (not necessarily matching git describe)
+  debian_epoch=""
+  if [ "${DEB_PACKAGE}" = "mender-client" ]; then
+    debian_epoch="1:"
+  fi
   debian_suffix="1+$OS_DISTRO+$OS_CODENAME"
   if [ "$VERSION" != "master" ] && git describe --tags --exact-match 2>/dev/null; then
     DEB_VERSION="$(git describe --tags --exact-match)"
     DEB_VERSION_NO_REV="${DEB_VERSION}"
     DEB_VERSION="${DEB_VERSION}-${debian_suffix}"
+    DEB_VERSION_NO_EPOCH="${DEB_VERSION}"
+    DEB_VERSION="${debian_epoch}${DEB_VERSION}"
   else
     DEB_VERSION="$(git tag | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' | sort -rV | head -n1)"
     if [ -z "${DEB_VERSION}" ]; then
@@ -115,6 +125,8 @@ get_deb_version() {
     # this has to have something to do with the `b` being a prefix
     # to binary numbers in perl
     DEB_VERSION="${DEB_VERSION}+builder${BUILD_ID}"
+    DEB_VERSION_NO_EPOCH="${DEB_VERSION}"
+    DEB_VERSION="${debian_epoch}${DEB_VERSION}"
   fi
 }
 
@@ -345,7 +357,7 @@ copy_deb_packages() {
     cp ${file} /output
   done
   # Echo the package version to /output
-  echo ${DEB_VERSION} > /output/${DEB_PACKAGE}-deb-version
+  echo ${DEB_VERSION_NO_EPOCH} > /output/${DEB_PACKAGE}-deb-version
   # Give packages same owner as the folder.
   chown --reference /output /output/*
 }
--- a/recipes/mender-client/debian-3.5.x/control
+++ b/recipes/mender-client/debian-3.5.x/control
@@ -6,7 +6,9 @@ Standards-Version: 3.9.2
 Build-Depends: debhelper (>= 9)
 
 Package: mender-client
-Conflicts: mender
+Conflicts: mender, mender-auth, mender-update, mender-snapshot, mender-setup
+Replaces: mender-auth, mender-update, mender-snapshot, mender-setup
+Provides: mender-auth, mender-update, mender-snapshot, mender-setup
 Architecture: any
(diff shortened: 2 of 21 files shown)


lluiscampos:
Build with `mender-client` 3.5.2 @ [Pipeline 1143084536](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/1143084536)

lluiscampos:
* Build with `mender-client` 3.5.2 @ [Pipeline 1143246012](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/1143246012)
* master build @ [Pipeline 1143240266](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/1143240266)

tranchitella (approved):

recipes/mender-client/debian-3.5.x/preinst
```diff
@@ -0,0 +1,64 @@
...
+                dpkg-divert --add --no-rename /usr/share/mender/inventory/mender-inventory-network
+                dpkg-divert --add --no-rename /usr/share/mender/inventory/mender-inventory-os
+                dpkg-divert --add --no-rename /usr/share/mender/inventory/mender-inventory-provides
+                dpkg-divert --add --no-rename /usr/share/mender/inventory/mender-inventory-rootfs-type
+                dpkg-divert --add --no-rename /usr/share/mender/inventory/mender-inventory-update-modules
+                dpkg-divert --add --no-rename /etc/mender/scripts/version
+                # stop and disable the v4 services
+                if systemctl is-active mender-authd; then
+                    systemctl stop mender-authd
+                    systemctl disable mender-authd
```
kacf, line 45: I think we need to remove these packages, we can't just disable them. Otherwise the system will be left with broken zombie packages afterwards, that have some, but not all of the files.

In fact, I'm wondering if removing should be done *instead of* using dpkg-divert, because the ordering is opposite here, so the files should already be gone by the time mender-client 3.5.x tries to put them in place.

tests/test_install_mender_sh.py
```diff
@@ -245,6 +261,123 @@ def test_upgrade_all_packages(
...
+    ):
+        # Install Mender 3.7.0 software with mender-client 3.5.2
+        local_apt_repo_from_upstream_packages(
+            generic_debian_container,
+            [
+                "m/mender-client/mender-client_3.5.2-1+debian+buster_amd64.deb",
+                "m/mender-connect/mender-connect_2.2.0-1+debian+buster_amd64.deb",
+                "m/mender-configure/mender-configure_1.1.1-1+debian+buster_all.deb",
+            ],
+            "/release_3_7_0",
```
kacf, line 265: Not against it, but in the spirit of future splitting, maybe I would omit 3.7 references here, and only use 3.x and 4.x client references instead.

Same with this other one, and with the comments.

kacf (commented):
Nice work!

tests/helpers.py
```diff
@@ -58,3 +60,14 @@ def upload_deb_package(
     ssh_connection.put(
         package_filename_path(package_version, package_name, package_arch)
     )
+
+
+def check_installed(conn, pkg, installed=True):
```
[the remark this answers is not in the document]
lluiscampos: Yes, that could be a nice improvement. I only moved the existing function in this commit so I let it to someone else to improve further :smiley:

mender-deb-package
```diff
@@ -87,16 +87,26 @@ get_deb_distribution() {
...
 get_deb_version() {
   # Create a version from Git.
-  #  - For Git tags: X.Y.Z-<debian_suffix>
-  #  - For master: X.Y.Z~git<commit-date>.<commit-sha>-<debian_suffix>+b<BUILD_ID>
+  #  - For Git tags: <debian-epoch>X.Y.Z-<debian_suffix>
+  #  - For master: <debian-epoch>X.Y.Z~git<commit-date>.<commit-sha>-<debian_suffix>+b<BUILD_ID>
   #     where X.Y.Z is latest tag (not necessarily matching git describe)
+  debian_prefix=""
+  if [ "${DEB_PACKAGE}" = "mender-client" ]; then
+    debian_prefix="1:"
```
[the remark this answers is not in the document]
lluiscampos: Yes, thanks. I should be as explicit as possible on what is this :+1:

recipes/mender-client/debian-3.5.x/control
```diff
@@ -6,7 +6,9 @@ Standards-Version: 3.9.2
 Build-Depends: debhelper (>= 9)
 
 Package: mender-client
-Conflicts: mender
+Conflicts: mender, mender-auth, mender-update, mender-flash, mender-snapshot, mender-setup
+Replaces: mender-auth, mender-update, mender-flash, mender-snapshot, mender-setup
+Provides: mender-auth, mender-update, mender-flash, mender-snapshot, mender-setup
```
kacf, line 11: I think `mender-flash` should be removed. This is a separate tool, and `autoremove` might still remove it later.

Also, mender-binary-delta uses mender-flash independently of mender-client now (for UBI), so this conflicts with that. Granted, I don't think you'll ever have mender-binary-delta on a system which also uses apt, but you can in principle.

kacf (approved):
One improvement below, but it looks good otherwise!

lluiscampos:
Builds:
* master [here](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/1144321511)
* 3.5.2 [here](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/1144323017)

recipes/mender-client/debian-3.5.x/control
lluiscampos: Makes sense.

As discussed privately, I will amend after the CI :green_circle: but merge without a new run. Then I will run again CI on the master build before publishing.

lluiscampos:
> Builds:
> 
>     * master [here](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/1144321511)
> 
>     * 3.5.2 [here](https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/pipelines/1144323017)

The failing tests are due to the APT repos being offline ([here](https://hub.mender.io/t/update-and-statement-on-the-mender-client-4-0-debian-package-incident/6497#update-1-2024-01-19-1000-utc-6)). The upgrade tests written in this pull requests are :evergreen_tree:

PR #406, opened 2024-01-22
Update `production` branch from `master`

lluiscampos:
Mainly bringing #405 to production

### DIFF
--- a/mender-deb-package
+++ b/mender-deb-package
@@ -87,16 +87,26 @@ get_deb_distribution() {
   fi
 }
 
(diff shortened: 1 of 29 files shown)


lluiscampos:
> as it is just a sync to prod I just read through it.
> 
> > Mainly bringing #405 to production
> 
> this is widely not true, #405 is tiny, this is bigger, unless I am missing something.

The updates in `production` branch trigger a deployment of `scripts/install-mender.sh` into get.mender.io. The changes in the other files are irrelevant.

PR #409, opened 2024-01-26
MEN-6971 part 1: Split out script tests

lluiscampos:

PR #410, opened 2024-01-31
MEN-6976: Add new package `mender-client4`

lluiscampos:


tests/helpers.py
```diff
@@ -59,11 +59,19 @@ def upload_deb_package(
...
-    conn."""
+    """Check whether the given package is installed on the device given by conn.
+    Check the specific dpkg Status to differentiate between installed (install ok installed)
+    and other status like removed but not purged (deinstall ok config-files)"""
 
     res = conn.run(f"dpkg --status {pkg}", warn=True)
     if isinstance(res, FabricResult):
-        assert (res.return_code == 0) == installed
+        retcode = res.return_code
+        output = res.stdout
```
kacf, line 69: Is this Python 2..? :open_mouth:

mender-deb-package
```diff
@@ -97,7 +97,7 @@ get_deb_version() {
   #  - For master: <debian-epoch>X.Y.Z~git<commit-date>.<commit-sha>-<debian_suffix>+b<BUILD_ID>
   #     where X.Y.Z is latest tag (not necessarily matching git describe)
   debian_epoch=""
-  if [ "${DEB_PACKAGE}" = "mender-client" ]; then
+  if [[ "${DEB_PACKAGE}" =~ ^mender-client(4)?$ ]]; then
     debian_epoch="1:"
```
kacf, line 101: So we continue mender-client4 at epoch 1. Can't we keep that only for mender-client? My impression is that it should be used sparingly.

recipes/mender-client4/debian-master/control
```diff
@@ -7,37 +7,36 @@ Build-Depends-Indep: cmake, debhelper (>= 10)
 Build-Depends: liblmdb-dev (>= 0.9.7), libc6-dev (>= 2.14), libboost-dev (>= 1.67.0), libboost-filesystem-dev (>= 1.67.0), libboost-log-dev (>= 1.67.0), libboost-regex-dev (>= 1.67.0), libboost-chrono-dev (>= 1.67.0), libboost-thread-dev (>= 1.67.0), libboost-atomic-dev (>= 1.67.0),  libssl-dev (>= 1.1.1), libarchive-dev (>= 3.2.2), libdbus-1-dev, libsystemd-dev
 Homepage: https://mender.io
 
-Package: mender-client
-Conflicts: mender
+Package: mender-client4
+Conflicts: mender, mender-client
+Replaces: mender-client, mender-auth, mender-update, mender-snapshot, mender-setup
```
kacf, line 12: Does it replace mender-update, mender-auth, mender-setup and mender-snapshot? This doesn't seem right...

oleorhagen (approved):
Very nice!

Tests were a treat :candy:

.gitlab-ci.yml
```diff
@@ -99,7 +99,11 @@ build:orig:debian:buster:amd64:
     - echo $CI_REGISTRY_PASSWORD | docker login -u $CI_REGISTRY_USER $CI_REGISTRY --password-stdin
   script:
     - apk --update --no-cache add bash
-    - ./docker-build-package $DISTRO $RELEASE $ARCH mender-client $MENDER_VERSION true
+    - if [[ "${MENDER_VERSION}" =~ ^3.* ]]; then
```
kacf, line 102: Pedantic: Should be `^3\..*`

.gitlab-ci.yml
```diff
@@ -123,7 +127,11 @@ build:packages:
     - echo $CI_REGISTRY_PASSWORD | docker login -u $CI_REGISTRY_USER $CI_REGISTRY --password-stdin
   script:
     - apk --update --no-cache add bash
-    - ./docker-build-package $DISTRO $RELEASE $ARCH mender-client $MENDER_VERSION
+    - if [[ "${MENDER_VERSION}" =~ ^3.* ]]; then
```
kacf, line 130: Pedantic: And here.

tests/test_install_mender_sh.py
```diff
@@ -225,25 +225,114 @@ def test_configure(
...
+
+        # MEN-7010
+        # Currently, production mender-configure still depends (only) on mender-client.
+        # We need first to update this one to the freshly built in order to make the rest of the
+        # test succeed. This part can be removed after MEN-7010 is finished.
+        generic_debian_container.run("apt install --assume-yes mender-configure")
+
+        # Now install freshly built mender-client4
+        generic_debian_container.run(
+            "DEBIAN_FRONTEND=noninteractive apt install --assume-yes mender-update mender-client4"
```
kacf, line 270: So it works fully automatic! Awesome!

kacf, line 270: You specified `mender-update` here. Intentional? Or should it be only `mender-client4`, to improve user experience?

I see you've done it differently below, but I don't quite see the intention behind it.

tests/test_install_mender_sh.py
```diff
@@ -225,25 +225,114 @@ def test_configure(
...
+        )
+
+        # mender-client should be removed and mender-client4 + all packages should be installed
+        check_installed(generic_debian_container, "mender-client", installed=False)
+        check_installed(generic_debian_container, "mender-client4")
         check_installed(generic_debian_container, "mender-update")
         check_installed(generic_debian_container, "mender-auth")
         check_installed(generic_debian_container, "mender-flash")
         check_installed(generic_debian_container, "mender-setup")
         check_installed(generic_debian_container, "mender-snapshot")
```
kacf, line 306: What about mender-connect and mender-configure?

kacf (approved):
Nice work! Just pedantic nitpicks and questions from me. 🙂

mender-deb-package
```diff
@@ -97,7 +97,7 @@ get_deb_version() {
   #  - For master: <debian-epoch>X.Y.Z~git<commit-date>.<commit-sha>-<debian_suffix>+b<BUILD_ID>
   #     where X.Y.Z is latest tag (not necessarily matching git describe)
   debian_epoch=""
-  if [ "${DEB_PACKAGE}" = "mender-client" ]; then
+  if [[ "${DEB_PACKAGE}" =~ ^mender-client(4)?$ ]]; then
```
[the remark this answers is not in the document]
lluiscampos: I don't know what I was thinking, `=~ ^mender-client4?$` should work the same way.

(I will likely remove it based on Kristian's comment)

tests/test_install_mender_sh.py
```diff
@@ -225,25 +225,114 @@ def test_configure(
...
+        # Now upgrade to the freshly built packages
         local_apt_repo_from_built_packages(generic_debian_container)
-        generic_debian_container.run("apt -y upgrade")
+        generic_debian_container.run("apt --assume-yes upgrade")
 
-        # All packages should be upgraded
+        # Only mender-client should be upgraded
         check_installed(generic_debian_container, "mender-client")
+        check_installed(generic_debian_container, "mender-client4", installed=False)
+        check_installed(generic_debian_container, "mender-update", installed=False)
```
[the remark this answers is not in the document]
lluiscampos: Yes :smile:

tests/helpers.py
lluiscampos: I don't understand. What exactly is Python 2, the `isinstance`?

mender-deb-package
lluiscampos: Mmm. I first thought that we had to keep it in order to upgrade but now I see that it doesn't make sense: it is a completely new package.

If we ever want to introduce a `mender-client` meta-package version 4 to force the upgrade then we need to have the epoch there, but as you are pointing out that doesn't imply that we need the epoch in `mender-client4`.

Thanks! I'll remove it.

recipes/mender-client4/debian-master/control
lluiscampos: The goal of this pacakge is to replace `mender-client`. However when we modified the package for the upgrade from 4.0.0 to 1:3.5.2, we introduce all these packages as "provides". See [here](https://github.com/mendersoftware/mender-dist-packages/commit/4a4af5c75bf1d5ba0c3d577a7e97d31706c0f283#diff-8ab54d0e6396495abbd8349b3f4bf813b79a93d881f3c4a2d1eca1930895e7c1R11).

I think we need to live with this, I tried many things and couldn't make it work otherwise.

tests/test_install_mender_sh.py
lluiscampos: Well, almost :sweat_smile: - See reply in comment below.

tests/helpers.py
kacf: I'm talking about the missing `decode`. But is it just a difference in Fabric maybe? I guess I'm not really clear on when this is a `FabricResult` and when it isn't.

tests/test_install_mender_sh.py
lluiscampos: This also was a painful discovery. If you have installed a package that depends on `mender-client | something`. That _something_ needs to be installed explicitly (or with a direct dependency, like `mender-update` depends on `mender-auth`) for the package to be kept in the system. In other words if you install only `mender-client4` apt will refuse to install it unless you remove `mender-connect` or other addons depending on `mender-client` (which has to be removed if you install `mender-client4`).

This is why the below test (without addons) does not require the `mender-update` explicit installation.

tests/test_install_mender_sh.py
lluiscampos: This test is named `test_upgrade_mender_v4_series_meta_package_only_client`, so the point is to verify the upgrade path with no addons.

mender-deb-package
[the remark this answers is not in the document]
lluiscampos: Reverted

tests/helpers.py
lluiscampos: Right. Yes, so `FabricResult` is using strings. The other case is a result from `subprocess.check_output`. Hopefully at some point we will remove fabric also from this test framework :)

tests/test_install_mender_sh.py
kacf: Ok, this is hard to wrap my head around. Let me try to recap what you said:

* Addons installed
    * `apt install mender-client4` will not work.
    * `apt install mender-client4 mender-update` will work.
* Addons not installed
    * `apt install mender-client4` will work.
    * `apt install mender-client4 mender-update` will work.

Bah! It's so annoying that the *addons* are what is breaking this... But I understand that we've run out of options here.

recipes/mender-client4/debian-master/control
lluiscampos: Mmmm. I think I tried this manually (an automated test for this is very hard) but maybe not in the last iteration. I will verify this again.

tests/test_install_mender_sh.py
kacf: Right, I missed that.

recipes/mender-client4/debian-master/control
kacf: > an automated test for this is very hard

Is it? Why?

I think `apt install mender-update mender-client4` is acceptable for the major upgrade, but I think it would be bad if you need to keep upgrading it like that also afterwards.

tests/test_install_mender_sh.py
lluiscampos: That is correct. For reference, see this extract from the log from `apt install` in the addons case:
```
...
dpkg: mender-client: dependency problems, but removing anyway as you requested:
 mender-connect depends on mender-client (>= 3.2.0) | mender-auth; however:
  Package mender-client is to be removed.
  Package mender-auth is not installed.
  Package mender-client which provides mender-auth is to be removed.
 mender-configure depends on mender-update | mender-client (>= 2.5.0); however:
  Package mender-update is not installed.
  Package mender-client which provides mender-update is to be removed.
  Package mender-client is to be removed.
 mender-connect depends on mender-client (>= 3.2.0) | mender-auth; however:
  Package mender-client is to be removed.
  Package mender-auth is not installed.
  Package mender-client which provides mender-auth is to be removed.
 mender-configure depends on mender-update | mender-client (>= 2.5.0); however:
  Package mender-update is not installed.
  Package mender-client which provides mender-update is to be removed.
  Package mender-client is to be removed.
...
```

recipes/mender-client4/debian-master/control
lluiscampos: > > an automated test for this is very hard
> 
> Is it? Why?
> 

Because you need two builds and we don't have an easy test framework for that. For example building `mender-client4` for tag `4.0.0` and for master. Then you first install the former and upgrade to the later.

> I think `apt install mender-update mender-client4` is acceptable for the major upgrade, but I think it would be bad if you need to keep upgrading it like that also afterwards.

I agree that that would be very bad. I however just verified that it works for `apt upgrade`:

```
vagrant@bullseye:~$ sudo apt upgrade
...
Preparing to unpack .../3-mender-auth_4.1.0~git20240206.f4459226-1+debian+buster+builderLOCAL_amd64.deb ...
Unpacking mender-auth (4.1.0~git20240206.f4459226-1+debian+buster+builderLOCAL) over (4.0.0-1+debian+buster) ...
Preparing to unpack .../4-mender-update_4.1.0~git20240206.f4459226-1+debian+buster+builderLOCAL_amd64.deb ...
Unpacking mender-update (4.1.0~git20240206.f4459226-1+debian+buster+builderLOCAL) over (4.0.0-1+debian+buster) ...
Preparing to unpack .../5-mender-client4_4.1.0~git20240206.f4459226-1+debian+buster+builderLOCAL_amd64.deb ...
Unpacking mender-client4 (4.1.0~git20240206.f4459226-1+debian+buster+builderLOCAL) over (4.0.0-1+debian+buster) ...
```

And for `apt install`:

```
vagrant@bullseye:~$ sudo apt install mender-client4
...
Preparing to unpack .../mender-client4_4.1.0~git20240206.f4459226-1+debian+buster+builderLOCAL_amd64.deb ...
Unpacking mender-client4 (4.1.0~git20240206.f4459226-1+debian+buster+builderLOCAL) over (4.0.0-1+debian+buster) ...
```

tests/test_install_mender_sh.py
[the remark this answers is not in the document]
lluiscampos: Oh, I see now that I didn't pull into this PR my other refactoring work (see [here](https://github.com/mendersoftware/mender-dist-packages/pull/409/commits/1a2e008a360159ea341ea0e73438c8fa238ad40b)). I hear you and I'll improve it further in the other value-add PR :pray:

recipes/mender-client4/debian-master/control
kacf: > Because you need two builds and we don't have an easy test framework for that. For example building `mender-client4` for tag `4.0.0` and for master. Then you first install the former and upgrade to the later.

My thinking was that you can just reinstall the same version. So not technically an upgrade. Granted, I'm not *certain* that this would trigger the same behavior in dpkg, but it's better than nothing, and should be pretty easy to test.

> I however just verified that it works for `apt upgrade`:

👍

tests/test_install_mender_sh.py
kacf: Got it. We'll have to live with this then. As long as subsequent upgrades will work, we're not too bad off.

recipes/mender-client4/debian-master/control
lluiscampos: > My thinking was that you can just reinstall the same version. So not technically an upgrade. Granted, I'm not _certain_ that this would trigger the same behavior in dpkg, but it's better than nothing, and should be pretty easy to test.
> 

Right, I didn't think about it. If it would have been an issue, I am not sure it would have shown with a `reinstall` command.

In any case it works (tested manually) and once this is in production it will be tested in regression with [this other test](https://github.com/mendersoftware/mender-dist-packages/blob/75a93c1cf33905488cba58a15709c54769fc55a9/tests/test_install_mender_sh.py#L228), so I think we are good here.

PR #413, opened 2024-02-01
Add libsystemd-dev as a build dependency for mender-client.

kacf:
Honestly I think this is a mistake by the upstream maintainers. Preumably it's because systemd is now providing dbus services, but this should not affect which package you need to depend on, at least not in a stable release.


lluiscampos (approved):

kacf:
Oh man, only now did I realize that this is of course because of this PR: https://github.com/mendersoftware/mender/pull/1463. Oh well, my judgmental comment about the Debian maintainers shall be in Git history forever! 😂

PR #418, opened 2024-02-12
chore: use specially purposed testing packages for the 4.0.0->v3 upgrade tests

tranchitella:


tests/test_install_mender_sh.py
```diff
@@ -337,6 +337,7 @@ def test_upgrade_mender_v4_series_explicit_auth_update(
         check_installed(generic_debian_container, "mender-configure")
 
 
+@pytest.mark.skip("skip until Mender 4.0.0 is properly released")
```
lluiscampos, line 340: The purpose of these tests were to verify the *upgrade* path from faulty v4 to epoch `1:` v3 with different variations. The upcoming release will publish different packages and these test won't run (as originally intended) with the new v4 packages.

For now, great to disable them and progress with the review.

However, for later, I think the best would be to save these packages *for tests purposes* somewhere in this repository and modify the tests, so that we keep testing in regression the upgrade to v3 path.

Note that we have now a [new set of tests](https://github.com/mendersoftware/mender-dist-packages/blob/650800c6b3495c323b98ed96f47a2c9f581dfd8d/tests/test_install_mender_sh.py#L228-L337) for the `mender-client4` package and it's relation with v3 `mender-client`.

tests/test_install_mender_sh.py
```diff
@@ -425,6 +426,7 @@ def test_upgrade_from_v3_to_build(
         check_installed(generic_debian_container, "mender-connect")
         check_installed(generic_debian_container, "mender-configure")
 
+    @pytest.mark.skip("skip until Mender 4.0.0 is properly released")
```
lluiscampos, line 429: This we don't need as we are already disabling the whole class.

lluiscampos (approved):
Feel free to merge this one without waiting for the pipeline - it won't add any value.

tests/test_install_mender_sh.py
lluiscampos: > (...) However, for later, I think the best would be to save these packages _for tests purposes_  (...)

I mean rescue them from a backup into this repo or some other test location

tranchitella: Reworked the PR to download the old files from a special location.

lluiscampos (approved):
Amazing. Thank you!

PR #419, opened 2024-02-14
feat: add support for mender-client4, mender-auth and mender-update components

tranchitella:
From now on, mender-client4 will be a default component in place of mender-client; however, the script falls back to mender-client (legacy, go client) for Ubuntu jammy and focal and for Debian bullseye and buster. As we don't support just yet other distributions, it means that the only way to install the Mender Client version 4 is to specify its components explicitly from the command line. In the future, when we'll support other distributions (e.g., bookworm), mender-client4 will be the default.

### DIFF
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -18,7 +18,9 @@ CHANNEL="stable"
 
 # All available components
 AVAILABLE_COMPONENTS="\
+mender-auth \
 mender-client \
+mender-client4 \
 mender-configure \
 mender-configure-demo \
 mender-configure-timezone \
@@ -26,10 +28,18 @@ mender-connect \
 mender-gateway \
 mender-monitor \
 mender-monitor-demo \
+mender-update \
 "
 
 # Default components (installed when no flags and no specified components)
 DEFAULT_COMPONENTS="\
+mender-client4 \
+mender-configure \
+mender-connect \
+"
+
+# Default components for legacy distributions (installed when no flags and no specified components)
+DEFAULT_COMPONENTS_LEGACY="\
 mender-client \
 mender-configure \
 mender-connect \
@@ -60,6 +70,9 @@ mender-monitor-demo \
 
 SELECTED_COMPONENTS="$DEFAULT_COMPONENTS"
 DEMO="0"
+
+# mender-setup CLI
+MENDER_SETUP_CLI="mender-setup"
 
 # Path where to install the Mender APT repository
 MENDER_APT_SOURCES_LIST="/etc/apt/sources.list.d/mender.list"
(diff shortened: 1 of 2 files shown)


tests/test_install_mender_sh.py
```diff
@@ -230,7 +230,7 @@ def test_configure(
             f"curl http://{SCRIPT_SERVER_ADDR}:{SCRIPT_SERVER_PORT}/install-mender.sh | bash -s -- mender-configure"
         )
 
-        check_installed(generic_debian_container, "mender-client")
+        check_installed(generic_debian_container, "mender-auth")
```
lluiscampos, line 233: `mender-configure` depends on `mender-update`, so it will be good to check for this one too :+1:

scripts/install-mender.sh
```diff
@@ -18,18 +18,28 @@ CHANNEL="stable"
...
 "
 
 # Default components (installed when no flags and no specified components)
 DEFAULT_COMPONENTS="\
+mender-client4 \
+mender-configure \
+mender-connect \
+"
+
+# Default components (installed when no flags and no specified components)
```
lluiscampos, line 41: Nit pick: modify a bit the description. For example
```suggestion
# Default components for legacy distributions (installed when no flags and no specified components)
```

lluiscampos (approved):
Nicely resolved!

lluiscampos:
@tranchitella See [my comment in the other related PR](https://github.com/mendersoftware/gui/pull/4317#discussion_r1490709197) - I am afraid we need to rework here a bit more.

tranchitella:
> @tranchitella See [my comment in the other related PR](https://github.com/mendersoftware/gui/pull/4317#discussion_r1490709197) - I am afraid we need to rework here a bit more.

see the latest commit

scripts/install-mender.sh
```diff
@@ -98,15 +117,22 @@ usage() {
...
-    echo "  -h, --help          print this help"
-    echo "  -c CHANNEL          channel: stable(default)|experimental"
-    echo "  --demo              use defaults appropriate for demo"
-    echo "  --commercial        install commercial components, requires --jwt-token"
-    echo "  --jwt-token TOKEN   Hosted Mender JWT token"
+    echo "  -h, --help             print this help"
+    echo "  -c CHANNEL             channel: stable(default)|experimental"
+    echo "  --demo                 use defaults appropriate for demo"
+    echo "  --commercial           install commercial components, requires --jwt-token"
+    echo "  --force-mender-client4 install the Mender Client 4.x serie on all the distros"
```
lluiscampos, line 124: Here and below: I think the right grammar is "series". Better to check with an English native :+1:

lluiscampos (approved):
Perfect! Thanks

PR #423, opened 2024-02-20
build armhf native

danielskinstad:


mender-deb-package
```diff
@@ -61,21 +61,28 @@ checkout_repo() {
...
   export PATH="$PATH:/usr/local/go/bin"
-  rm -vf "go$GOLANG_VERSION.linux-amd64.tar.gz"
+  rm -vf "go$GOLANG_VERSION.linux-${GOLANG_ARCH}.tar.gz"
 }
 
 get_os_version() {
   OS_DISTRO="$(. /etc/os-release && echo "$ID")"
+  if [ $OS_DISTRO = "raspbian" ]; then
+      OS_DISTRO="debian"
+  fi
```
lluiscampos, line 85: This is fine but it deserves a comment for the future you or the future me :nerd_face:

mender-deb-package
```diff
@@ -240,7 +230,11 @@ EOF
     rm -f debian/patches/series
   fi
 
+  apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 6ED0E7B82643E131 2> /dev/null
+  apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 0E98404D386FA1D9 2> /dev/null
```
lluiscampos, line 234: I guess you have gone back and forth with this already, but I cannot still understand why this cannot be done at built time.

mender-deb-package
```diff
@@ -240,7 +230,11 @@ EOF
...
+
   apt-get update
+  apt-get -y install cmake
```
lluiscampos, line 237: (this is good to stay here, by the way - my above comment is only for the `apt-key` commands.

mender-deb-package
lluiscampos: Also, `apt-key` is deprecated, we should use `trusted.gpg.d`. Google a bit on your side and see a similar fix we did some time ago here: https://github.com/mendersoftware/mender-dist-packages/commit/7d4341b218e9d62c5d9f4e4c0ed8a5d7605104f1

lluiscampos (changes requested):
Excellent. This one is quite straight forward.

See my comments for some suggestions and requests. Also, as agreed f2f, please verify that the new packages can be installed and executed on a real :pie:

lluiscampos (approved):

lluiscampos:
@danielskinstad Good to go :partying_face:

PR #424, opened 2024-02-20
chore: improve the autodetection of `mender-setup` vs. `mender setup`

tranchitella:
Changelog: None
Ticket: None


lluiscampos (approved):
I think we can simplify the help text, see below.

Pre-approving! Well done.

PR #427, opened 2024-02-22
test: New test to validate switch to `mender-client4` package

lluiscampos:


tranchitella (commented):
This test will always fail... not sure what you plan to do with this PR.

lluiscampos:
> This test will always fail... not sure what you plan to do with this PR.

Yes, I will close this one for now. Whenever I have the time for the refactoring of the tests (so that we can test other distros) I will pick this back :+1:

PR #430, opened 2024-02-25
Don't merge, test only: build with docker sock mount

oldgiova:
Using not privileged docker runner, with docker sock mount


oldgiova:
Test done

PR #433, opened 2024-03-21
fix: restart the correct systemd services when installing the Mender Client version 4

tranchitella:


scripts/install-mender.sh
```diff
@@ -466,7 +466,14 @@ do_setup_mender_client() {
...
     `mender_setup_cli` $MENDER_SETUP_ARGS
-    pidof systemd && systemctl restart mender-client
+    pidof systemd && (
+        if [[ "$SELECTED_COMPONENTS" == *"mender-client4"* ]]; then
+            systemctl restart mender-authd
+            systemctl restart mender-updated
+        else
+            systemctl restart mender-client
+        fi
+    )
```
merlin-northern, line 476:
```suggestion
    pidof systemd && {
        if [[ "$SELECTED_COMPONENTS" == *"mender-client4"* ]]; then
            systemctl restart mender-authd
            systemctl restart mender-updated
        else
            systemctl restart mender-client
        fi
    }
```
no need for subshell here. (unless it was on purpose)

merlin-northern (approved):
looks good. one comment up to you.

scripts/install-mender.sh
```diff
@@ -466,7 +466,14 @@ do_setup_mender_client() {
 
     echo "  Setting up mender with options: $MENDER_SETUP_ARGS"
     `mender_setup_cli` $MENDER_SETUP_ARGS
-    pidof systemd && systemctl restart mender-client
+    pidof systemd && (
+        if [[ "$SELECTED_COMPONENTS" == *"mender-client4"* ]]; then
```
lluiscampos, line 470: This won't work if, for example, I "select" `mender-update` for install by running `get.mender-io ... mender-update`. (Didn't we fix this elsewhere?)

We should instead check for all the services individually. What about:
```
systemctl status mender-client && systemctl restart mender-client
systemctl status mender-updated && systemctl restart mender-updated
systemctl status mender-authd && systemctl restart mender-authd
```

lluiscampos (changes requested):

scripts/install-mender.sh
tranchitella: @lluiscampos please properly check the function: you reach this point only if you are installing mender-client or mender-client4.

scripts/install-mender.sh
tranchitella: thanks!

scripts/install-mender.sh
lluiscampos: Thanks for pointing that out,

Then, it is that [previous check](https://github.com/mendersoftware/mender-dist-packages/pull/433/files#diff-4f0df1d73eedfab863e987b0c7f29079133c80b39e5e68b15eee877095dc612fR458) that will misbehave in the scenario that I was describing: if I install `mender-update` directly, it will pull `mender-setup` (recommended package) and then this function should still run if I have specified setup arguments.

tranchitella: ok, let's do it that way

lluiscampos (approved):

scripts/install-mender.sh
```diff
@@ -466,7 +466,11 @@ do_setup_mender_client() {
 
     echo "  Setting up mender with options: $MENDER_SETUP_ARGS"
     `mender_setup_cli` $MENDER_SETUP_ARGS
-    pidof systemd && systemctl restart mender-client
+    pidof systemd && {
+        systemctl status mender-authd >/dev/null 2>&1 && systemctl restart mender-authd
+        systemctl status mender-updated >/dev/null 2>&1 && systemctl restart mender-updated
+        systemctl status mender-client >/dev/null 2>&1 && systemctl restart mender-client
```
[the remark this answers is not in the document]
tranchitella: ok, let's try a different approach

merlin-northern (approved):
looks good. thank you for taking this.

PR #449, opened 2024-04-12
test: building on a virtual arm server

oldgiova:
Ticket: INF-3478

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -69,10 +69,20 @@ variables:
     matrix:
       - DISTRO: debian
         RELEASE: [buster, bullseye, bookworm]
+        ARCH: [amd64, arm64]
+      - DISTRO: debian # cannot build this with arm64 host
+        RELEASE: [buster]
         ARCH: [amd64, armhf, arm64]
       - DISTRO: ubuntu
         RELEASE: [focal, jammy]
         ARCH: [amd64, armhf, arm64]
+
+.mender-dist-packages-image-matrix-armhf:
+  parallel:
+    matrix:
+      - DISTRO: debian
+        RELEASE: [bullseye, bookworm]
+        ARCH: [armhf]
 
 include:
   - project: 'Northern.tech/Mender/mendertesting'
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -69,11 +69,18 @@ variables:
...
       - DISTRO: ubuntu
         RELEASE: [focal, jammy]
         ARCH: [amd64, armhf, arm64]
 
+.mender-dist-packages-image-matrix-arm:
+  parallel:
+    matrix:
+      - DISTRO: debian
+        RELEASE: [buster, bullseye, bookworm]
+        ARCH: [armhf, arm64]
```
lluiscampos, line 82: This looks really promising! Great idea introducing ARM CI runners.

The only jobs that we need to run on actual ARM machines are:
```
    matrix:
      - DISTRO: debian
        RELEASE: [buster, bullseye, bookworm]
        ARCH: [armhf]
```

The rest of the ARM combinations, for example `ubuntu armhf`, `ubuntu arm64` and `debian arm64` are using the regular cross-compile toolchain from Debian upstream and they need to run in native x86_64 (we could experiment on having these also doing native builds on ARM, but let's keep that for a different iteration).

oldgiova:
> @oldgiova I cannot make sense of the build failure.
> 
> For comparison, I've been looking at the `build:packages:armhf: [debian, buster, armhf] ` build for:
> 
> * Your PR: https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/jobs/6634947099
> * Nightly from few days ago: https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/jobs/6709951733
> 
> And as far as I can see: the relevant source is the same and the builder image is the same (search for `sha256:c12c49f14e030a...` (which, indirectly, means that the `cmake` version used is the same). The build error is related to a missing file in the filesystem, which just doesn't make sense to me as this is run on a container in both cases.
> 
> As I stated privately, we are sun-setting this OS version so the problem will be avoided soon but in any case we should get a better understanding of what is going on here to be confident that won't affect other OS versions 🤔

I tried with different build flags, still same issue. I suspect some issue in the building tools/libraries that have been fixed in the next Debian distros.
Since we're about to deprecate buster, it makes no sense to continue on this.
Moving on arm64 only bullseye and bookworn

lluiscampos (approved):
> (...). Moving on arm64 only bullseye and bookworn

I agree with the conclusion :+1:

lluiscampos:
@oldgiova FYI I updated your PR from latest master and, indirectly, dropped one commit of yours that was already merged in master [here](https://github.com/mendersoftware/mender-dist-packages/commit/7741987ef0afe7baed608675f6686255fca1fbc3).

PR #459, opened 2024-05-23
chore: sort get.mender.io script distribution output

danielskinstad:
Detected distribution is now displayed after the mender banner

Ticket: MEN-7274

### DIFF
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -319,6 +319,15 @@ init() {
 
     parse_args "$@"
 
+    echo "  Detected distribution:"
+    printf "\t%s/%s\n" "$LSB_DIST" "$DIST_VERSION"
+
+    if [[ "$LSB_DIST" == "raspbian" ]]; then
+        LSB_DIST="debian"
+        echo "  Raspbian detected. Using compatible distribution:"
+        printf "\t%s/%s\n" "$LSB_DIST" "$DIST_VERSION"
+    fi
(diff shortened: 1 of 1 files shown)


scripts/install-mender.sh
```diff
@@ -319,6 +319,15 @@ init() {
...
     parse_args "$@"
 
+    echo "  Detected distribution:"
+    printf "\t%s/%s\n" "$LSB_DIST" "$DIST_VERSION"
+
+    if [[ "$LSB_DIST" == "raspbian" ]]; then
+        LSB_DIST="debian"
+        echo "  Raspbian detected. Using compatible distribution:"
+        printf "\t%s/%s\n" "$LSB_DIST" "$DIST_VERSION"
+    fi
```
lluiscampos, line 329: While putting this in `init` resolves the issue, I think this special treatment of `raspbian` is out of context. I think it should be as close as possible to where `LSB_DIST` is originally defined (iow `check_dist_and_version` function). What do you think?

I think this is as simple as swapping the order of the functions in the "main" below, like:
```
banner
check_dist_and_version
init "$@"
print_components
...
```

Or is there something else that would break with that approach?

tranchitella:
I believe this is a simpler solution:
https://github.com/mendersoftware/mender-dist-packages/pull/463

tranchitella (changes requested):
See my proposal

PR #463, opened 2024-05-27
fix: sort output when running install-mender.sh

tranchitella:
Ticket: MEN-7274
Changelog: None


lluiscampos (approved):
[Exactly](https://github.com/mendersoftware/mender-dist-packages/pull/459#discussion_r1613190137). Thanks :+1:

PR #462, opened 2024-05-26
fix: install gcc in tests:acceptance

merlin-northern:
Changelog: Title
Ticket: None

PR #464, opened 2024-05-27
feat: always install mender-client4 if the channel is `experimental`

tranchitella:
Changelog: None
Ticket: MEN-7284

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -227,11 +227,12 @@ test:check-python3-formatting:
     - git submodule update --init --recursive
     # Log in to pull test image from registry.gitlab.com
     - echo $CI_REGISTRY_PASSWORD | docker login -u $CI_REGISTRY_USER $CI_REGISTRY --password-stdin
-  script:
+    # Install dependencies
     - apk --update --no-cache add python3 py3-pip gcc openssh-client make openssl-dev
       libffi-dev libc-dev python3-dev bash musl-dev rust cargo
+    - pip3 install --break-system-packages -r tests/requirements.txt
+  script:
     - cd tests
-    - pip3 install --break-system-packages -r requirements.txt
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -292,7 +292,7 @@ test:acceptance:script:
   extends: .test:acceptance
   needs: []
   script:
-    - apk --update --no-cache add python3 py3-pip
+    - apk --update --no-cache add python3 py3-pip gcc make openssh-client openssl-dev python3-dev musl-dev libffi-dev
```
lluiscampos, line 295: See my review at https://github.com/mendersoftware/mender-dist-packages/pull/462#pullrequestreview-2080156858

tranchitella: @lluiscampos ok, I applied your changes.

scripts/install-mender.sh
```diff
@@ -15,6 +15,9 @@
 set -e
 
 CHANNEL="stable"
+if [[ "$*" == *"experimental"* ]]; then
+    CHANNEL="experimental"
+fi
```
lluiscampos, line 20: This defeats `parse_args` function. Instead, isn't it possible to change the order of the main calls so that we do `init` before `check_dist_and_version`?

This would break for example if in the future we have a `something-experimental` package or `experimental` string appearing in some other way in the CLI args.

scripts/tests/test_install_mender_sh.py
```diff
@@ -175,14 +175,18 @@ def test_default(
     ):
         """Default, no arg install installs mender-client and add-ons (stable)."""
 
+        client_package = (
+            "mender-client4" if channel == "experimental" else "mender-client"
+        )
```
lluiscampos, line 180: :+1:

lluiscampos (approved):

PR #470, opened 2024-06-07
chore: save additional information in the /incoming flag

merlin-northern:
Ticket: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -315,10 +315,15 @@ test:acceptance:script:
     - apt update && apt install -yyq awscli
     # Lock the bucket to block concurrent jobs
     - while aws s3 ls s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/lock; do
-    -   echo "$S3_BUCKET_REPO_PATH locked, waiting..."
+    -   rm -f /tmp/lock.in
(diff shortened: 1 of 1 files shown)


merlin-northern:
I am a bit unsure if that will work with the output redirection -- it never worked for me, soe @lluiscampos just feel free to destroy this idea.

.gitlab-ci.yml
```diff
@@ -315,10 +315,15 @@ test:acceptance:script:
     - apt update && apt install -yyq awscli
     # Lock the bucket to block concurrent jobs
     - while aws s3 ls s3://$S3_BUCKET_NAME/$S3_BUCKET_REPO_PATH/lock; do
-    -   echo "$S3_BUCKET_REPO_PATH locked, waiting..."
+    -   rm -f /tmp/lock
```
lluiscampos, line 318: This line is unnecessary

lluiscampos (approved):

.gitlab-ci.yml
merlin-northern: woops. nice catch! it was supposed to be `/tmp/lock.in` . fixed.

merlin-northern: I added one more removal and an echo with waiting, which was there, so I did not try to smuggle too much 😘

PR #476, opened 2024-07-01
ci: Version pin docker to 26

alfrunes:


.gitlab-ci.yml
```diff
@@ -192,7 +192,9 @@ test:check-python3-formatting:
 
 .test:acceptance:
   stage: test
-  image: docker:dind
+  image: docker:26-cli
+  services:
+    - docker:26-dind
```
lluiscampos, line 197: Docker in docker as a service won't work because these tests do SSH into the virtual device container.

lluiscampos (commented):
It seems to have fixed the issue at hand, thanks! But wrt the acceptance tests, I think we need to revert or find another kind of fix. See comment below.

.gitlab-ci.yml
alfrunes: Deleted the last commit.

PR #489, opened 2024-09-18
ci: using direct subgroup dependency proxy

oldgiova:
To allow users without top group permissions to run jobs

Ticket: None


danielskinstad (approved):
Looks good to me, thanks!

PR #492, opened 2024-10-26
perf: speed up aws cli by ignoring imds

oldgiova:
IMDS is an api service available within AWS. AWS cli by default is trying to get metadata informations from the address 169.254.169.254, which is the internal IMDS endpoint. Our pipelines run in Hetzner and GCP, where this service is not available, and every aws cli command time out for about 5, slowing down the pipelines in the long run.

Ticket: QA-751


lluiscampos (approved):
Great finding :rocket:

PR #496, opened 2024-11-20
MEN-7731: Build packages for Debian armhf (but no publish)

lluiscampos:
Use the new image names from `mender-test-containers` to build both the Raspberry Pi OS armhf packages (previously known as Debian armhf) and actual Debian armhf packages.

Note that the publishing must be prevented until we re-design the APT repository structure.


danielskinstad (approved):
Looks good to me

PR #505, opened 2025-01-15
MEN-7930: Remove Ubuntu Focal 20.04 deb packages

lluiscampos:


lluiscampos:
> Can probably remove [this](https://github.com/lluiscampos/mender-dist-packages/blob/b4f18993a06684162016892d7de809d51a612fd9/scripts/install-mender.sh#L565) as well?

I thought about it, but I decided to leave it for the same reason that we did not remove [this](https://github.com/lluiscampos/mender-dist-packages/blob/b4f18993a06684162016892d7de809d51a612fd9/scripts/install-mender.sh#L586) one. The reason being that although we won't be updating any longer the packages there, we are not removing the repository neither, so in theory someone could still use the script to install the repo.

PR #508, opened 2025-02-13
ci: docker login retry

oldgiova:
From time to time gitlab.com is timing out when issuing docker login, this change allow for a retry

This should solve this case: https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/jobs/9126549483

Ticket: QA-921

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -95,7 +95,7 @@ variables:
 
 .dind-login: &dind-login
   - mkdir -p $HOME/.docker && echo $DOCKER_AUTH_CONFIG > $HOME/.docker/config.json
-  - docker login --username $CI_REGISTRY_USER --password $CI_REGISTRY_PASSWORD $CI_REGISTRY
+  - for retry in 1 2 3 4 5; do docker login --username $CI_REGISTRY_USER --password $CI_REGISTRY_PASSWORD $CI_REGISTRY && break || sleep $retry; done
 
 include:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -95,7 +95,7 @@ variables:
 
 .dind-login: &dind-login
   - mkdir -p $HOME/.docker && echo $DOCKER_AUTH_CONFIG > $HOME/.docker/config.json
-  - docker login --username $CI_REGISTRY_USER --password $CI_REGISTRY_PASSWORD $CI_REGISTRY
+  - for retry in {1..5}; do docker login --username $CI_REGISTRY_USER --password $CI_REGISTRY_PASSWORD $CI_REGISTRY && break || sleep $retry; done
```
lluiscampos, line 98: Careful, this won't work on shell, it is a bash extension - we do use bash in here but I prefer to avoid these as we use to copy paste things around and soon or alter we mess it up :see_no_evil: 

Why not just:
```suggestion
  - for retry in 1 2 3 4 5; do docker login --username $CI_REGISTRY_USER --password $CI_REGISTRY_PASSWORD $CI_REGISTRY && break || sleep $retry; done
```

lluiscampos (approved):
Not very important, but see my comment below.

.gitlab-ci.yml
oldgiova: Smart suggestion! Thank you!

PR #518, opened 2025-04-02
ci: retry on docker not running.

merlin-northern:
Everyday we have a failure of this sort:

docker: error during connect: Head "https://docker:2376/_ping": dial tcp: lookup docker on 127.0.0.11:53: server misbehaving.

https://gitlab.com/Northern.tech/Mender/mender-dist-packages/-/jobs/9593744363

PoC type of approach (I do not know how to do retry conditionally): retry job on simple docker not running.

Changelog: Title
Ticket: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -73,6 +73,16 @@ variables:
   # Don't slow down aws cli commands with connections to IMDS: we're not in AWS
   AWS_EC2_METADATA_DISABLED: "true"
 
+include:
+  - project: 'Northern.tech/Mender/mendertesting'
+    file: '.gitlab-ci-check-commits-signoffs.yml'
+  - project: 'Northern.tech/Mender/mendertesting'
+    file: '.gitlab-ci-check-license.yml'
+  - project: 'Northern.tech/Mender/mendertesting'
+    file: '.gitlab-ci-check-python3-format.yml'
+  - project: 'Northern.tech/Mender/mendertesting'
+    file: '.gitlab-ci-github-status-updates.yml'
+
 # NOTE: To add distributions, modify first the matrix in mender-test-containers repository
 .mender-dist-packages-image-matrix-cross:
   parallel:
@@ -97,15 +107,29 @@ variables:
   - mkdir -p $HOME/.docker && echo $DOCKER_AUTH_CONFIG > $HOME/.docker/config.json
   - for retry in 1 2 3 4 5; do docker login --username $CI_REGISTRY_USER --password $CI_REGISTRY_PASSWORD $CI_REGISTRY && break || sleep $retry; done
 
-include:
-  - project: 'Northern.tech/Mender/mendertesting'
-    file: '.gitlab-ci-check-commits-signoffs.yml'
-  - project: 'Northern.tech/Mender/mendertesting'
-    file: '.gitlab-ci-check-license.yml'
-  - project: 'Northern.tech/Mender/mendertesting'
-    file: '.gitlab-ci-check-python3-format.yml'
-  - project: 'Northern.tech/Mender/mendertesting'
-    file: '.gitlab-ci-github-status-updates.yml'
+.requires-docker: &requires-docker
+  - DOCKER_RETRY_SLEEP_S=2
+  - DOCKER_RUNNING=false
+  - for try in 4 3 2 1; do
+  -  docker ps && DOCKER_RUNNING=true
+  -  if [ "${DOCKER_RUNNING}" == "true" ]; then
+  -   break
+  -  fi
+  -  sleep "${DOCKER_RETRY_SLEEP_S}"
+  - done
+  - if [ "${DOCKER_RUNNING}" != "true" ]; then
+  -  exit 192
+  - fi
+
+.build:base:
+  before_script:
+    - *dind-login
+    - *requires-docker
+  retry:
+    max: 2
+    exit_codes:
+      - 137
+      - 192
 
 stages:
   - build:orig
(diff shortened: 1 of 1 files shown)


danielskinstad:
We could perhaps use this https://docs.gitlab.com/ci/yaml/#retryexit_codes to retry on certain exit codes

merlin-northern:
> We could perhaps use this https://docs.gitlab.com/ci/yaml/#retryexit_codes to retry on certain exit codes

bloody amazin! thanks!

merlin-northern:
> > We could perhaps use this https://docs.gitlab.com/ci/yaml/#retryexit_codes to retry on certain exit codes
> 
> bloody amazin! thanks!

I have to correct one think, docker ps failing will trigger exit and failure earlier, I have to disable the shell opts.

.gitlab-ci.yml
```diff
@@ -189,6 +199,9 @@ build:packages:cross:
       - output/commercial
       - output/opensource
   parallel: !reference [.mender-dist-packages-image-matrix-cross, parallel]
+  retry:
+    max: 2
+    exit_codes: 192
```
oldgiova, line 204: Also the 137 exit code
```suggestion
  retry: 
    max: 2
    exit_codes: 
      - 137
      - 192
```

oldgiova (approved):
That's a great one! Only a nitpick but this is very valuable imho

merlin-northern:
> > > We could perhaps use this https://docs.gitlab.com/ci/yaml/#retryexit_codes to retry on certain exit codes
> > 
> > 
> > bloody amazin! thanks!
> 
> I have to correct one think, docker ps failing will trigger exit and failure earlier, I have to disable the shell opts.

okey checked. should be ok.

.gitlab-ci.yml
merlin-northern: fixed!
and thanks!

.gitlab-ci.yml
```diff
@@ -166,6 +166,20 @@ build:packages:cross:
...
+    -  docker ps && DOCKER_RUNNING=true
+    -  if [[ "${DOCKER_RUNNING}" == "true" ]]; then
+    -   break
+    -  fi
+    -  sleep "${DOCKER_RETRY_SLEEP_S}"
+    -  let DOCKER_RETRIES--
+    - done
+    - if [[ "${DOCKER_RUNNING}" != "true" ]]; then
+    -  exit 192
+    - fi
```
lluiscampos, line 182: Please put this somewhere else and reuse it in the other jobs:
* build:orig:debian:bullseye:amd64
* build:packages:sim

merlin-northern: actually I started all this with adding to the QA meeting agenda, I want to put it in some type of docker:required:template or something or this sort. we can test it in here, and then create a follow up ticket. and I will adhere to your comment. thank you.

merlin-northern: tried my best in the separate commit. what do you think?
also: `build:packages:sim` is already covered I think via `build:packages:cross:` right?

.gitlab-ci.yml
```diff
@@ -133,6 +149,7 @@ build:orig:debian:bullseye:amd64:
   before_script:
     - *dind-login
   script:
+    - *requires-docker
```
lluiscampos, line 152: Instead of having this and the retry exit codes in the two jobs, we could now have a base job like:

```
.build:base:
  before_script:
    - *dind-login
    - *requires-docker
  retry:
    max: 2
    exit_codes:
      - 137
      - 192
```

And then have both `build:orig:debian:bullseye:amd64` and `build:packages:cross` extend it.

.gitlab-ci.yml
```diff
@@ -107,6 +107,22 @@ include:
   - project: 'Northern.tech/Mender/mendertesting'
     file: '.gitlab-ci-github-status-updates.yml'
 
+.requires-docker: &requires-docker
```
lluiscampos, line 110: Nit pick, this should be next to the other yaml anchor `&dind-login`.

I think that I like more your porposal of having it after the `include`, but then both should be here.

.gitlab-ci.yml
```diff
@@ -107,6 +107,22 @@ include:
...
+  - DOCKER_RETRIES=4
+  - DOCKER_RETRY_SLEEP_S=1
+  - DOCKER_RUNNING=false
+  - while [[ $DOCKER_RETRIES -gt 0 ]]; do
```
lluiscampos, line 114: Can we keep the style of the existing snippet and do `for retry in 1 2 3 4 5`? Two reasons, 1) consistency with existing code and 2) do not add a bash dependency

lluiscampos (changes requested):
You are right about `build:packages:sim` :+1: - Thanks for checking

The two commits should be marked as `ci:`, they are no software fix. They could also be squashed, but that I don't have an strong opinion.

Also see my new comments below :point_down: 

Requesting changes for the commit formalities, the rest of the comments apply what you consider suitable.

.gitlab-ci.yml
merlin-northern: sure

.gitlab-ci.yml
merlin-northern: sure, no worries, moved around I hope I have not lost anything.

.gitlab-ci.yml
merlin-northern: sure. so, now we are only one step away from moving `.build:base:` to a template? :>

lluiscampos (approved):

lluiscampos:
Thank you :rocket:

PR #520, opened 2025-04-14
chore: Add RPM spec for mender-artifact

vpodzime:
The very first step towards providing RPM packages for our tools.

Ticket: none
Changelog: none

### DIFF
--- a/recipes/mender-artifact/rpm/mender-artifact.spec
+++ b/recipes/mender-artifact/rpm/mender-artifact.spec
@@ -1,0 +1,62 @@
+%define mender_artifact_version 4.1.0
+%define tag build1
+
+# We don't want debuginfo packages
+%global debug_package %{nil}
+
+Name: mender-artifact
+Summary: A CLI tool for working with Mender artifacts
+Version: %{mender_artifact_version}
+Release: 1%{?dist}
+License: Apache-2.0
+URL: https://github.com/mendersoftware/mender-artifact/
+
+# Once we have releases at GitHub not just tags, we should use a URL like the
+# one below, instead:
+# Source0: https://github.com/mendersoftware/mender-artifact/releases/download/...
+# The variant below, with the tarball name, results in a warning from rpmlint
+# that it's not a valid URL. But the URL to get it from looks like this:
+#   https://github.com/mendersoftware/mender-artifact/archive/refs/tags/4.1.0.tar.gz
+# which has no package name in tarball file name. So we use a custom file name
+# until we can use a release.
+Source0: %{name}-%{version}-%{tag}.tar.gz
+
+BuildRequires: git
+BuildRequires: golang
+BuildRequires: openssl-devel
+
+# Fedora 41 deprecated the Engine API [1] and thus a separate package needs to
+# be installed to get /usr/include/openssl/engine.h. Fedora 40 knows
+# openssl-devel provides that file and would not install anything extra with the
+# line below, but other RPM-based distros would have no clue.
+# [1] https://fedoraproject.org/wiki/Changes/OpensslDeprecateEngine
+%if %{?fedora}%{!?fedora:0} > 40
+BuildRequires: openssl-devel-engine
+%endif
+
+%description
+A CLI tool to work with a Mender artifact, which is a file that can be recognized
+by its .mender suffix. Mender artifacts can contain binaries, metadata, checksums,
+signatures and scripts that are used during a deployment.
+
+%prep
+%setup -q -n %{name}-%{version}
+
+%build
+make build
+
+%install
+install -m 0755 -vd %{buildroot}%{_bindir}
+# Install the binary ourselves, instead of `GOBIN=... make install` because that
+# does extra things we don't need/want.
+install -m 0755 -vp ./mender-artifact %{buildroot}%{_bindir}/
+
+%files
+%{_bindir}/mender-artifact
+%doc README.md
+%license LICENSE
+# TODO: Should we add artifact-format-v2.md artifact-format-v3.md?
+
+%changelog
+* Mon Apr 14 2025 Vratislav Podzimek <vratislav.podzimek@nothern.tech> - 4.1.0-1
+- Initial release


vpodzime:
spec used to build packages in https://software.opensuse.org//download.html?project=home%3Anorthern.tech%3AMender&package=mender-artifact

vpodzime:
It's hard to believe that I broke these by adding a file not used anywhere:

image

recipes/mender-artifact/rpm/mender-artifact.spec
```diff
@@ -0,0 +1,62 @@
+%define mender_artifact_version 4.1.0
```
lluiscampos, line 1: For debian, we have been versioning the recipes by minor releases. `mender-artifact` happens to not have introduced any changes, but if you look for example to [`mender-connect`](https://github.com/mendersoftware/mender-dist-packages/tree/master/recipes/mender-connect) or [`mender-client` (Go)](https://github.com/mendersoftware/mender-dist-packages/tree/master/recipes/mender-client) you well see what I mean. It is a weird design, but basically is like "use your minor series if exist, otherwise master".

The idea is that mender-dist-packages should be able to build any version of the packages, for example we might have two minor releases supported for some time for which packages need to be build while there is some change in the spec or debian rules.

This was a long intro to say:

I think this file should be moved into either:
* `recipes/mender-artifact/rpm-master/mender-artifact.spec` (this would be similar to the debian ones) or
* `recipes/mender-artifact/rpm/mender-artifact-master.spec` (taking advantage that rpm packaging is a single file)

recipes/mender-artifact/rpm/mender-artifact.spec
```diff
@@ -0,0 +1,62 @@
...
+
+# Once we have releases at GitHub not just tags, we should use a URL like the
+# one below, instead:
+# Source0: https://github.com/mendersoftware/mender-artifact/releases/download/...
+# The variant below, with the tarball name, results in a warning from rpmlint
+# that it's not a valid URL. But the URL to get it from looks like this:
+#   https://github.com/mendersoftware/mender-artifact/archive/refs/tags/4.1.0.tar.gz
+# which has no package name in tarball file name. So we use a custom file name
+# until we can use a release.
+Source0: %{name}-%{version}-%{tag}.tar.gz
```
lluiscampos, line 22: How will be build nightlies for the experimental repo? This is one of the point of high complexity in the debian recipes, where we have some crazy logic in the build script to generate valid versions out of Git info + CI pipeline info. See [here](https://github.com/mendersoftware/mender-dist-packages/blob/aa6c317b865dc019516df3f0187d80b26b2c43ec/mender-deb-package#L97).

recipes/mender-artifact/rpm/mender-artifact.spec
```diff
@@ -0,0 +1,62 @@
...
+install -m 0755 -vp ./mender-artifact %{buildroot}%{_bindir}/
+
+%files
+%{_bindir}/mender-artifact
+%doc README.md
+%license LICENSE
+# TODO: Should we add artifact-format-v2.md artifact-format-v3.md?
+
+%changelog
+* Mon Apr 14 2025 Vratislav Podzimek <vratislav.podzimek@nothern.tech> - 4.1.0-1
```
lluiscampos, line 61: Our changelogs in the existing package point to the online docs: `Release 1.2.3. See online docs for complete changelog`. Also the author should be mender@northern.tech

recipes/mender-artifact/rpm/mender-artifact.spec
```diff
@@ -0,0 +1,62 @@
...
+install -m 0755 -vd %{buildroot}%{_bindir}
+# Install the binary ourselves, instead of `GOBIN=... make install` because that
+# does extra things we don't need/want.
+install -m 0755 -vp ./mender-artifact %{buildroot}%{_bindir}/
+
+%files
+%{_bindir}/mender-artifact
+%doc README.md
+%license LICENSE
+# TODO: Should we add artifact-format-v2.md artifact-format-v3.md?
```
lluiscampos, line 58: That could be a nice addition, although I would do it with a dedicated task and do it for both debian and rpm packages.

recipes/mender-artifact/rpm/mender-artifact.spec
```diff
@@ -0,0 +1,62 @@
...
+%install
+install -m 0755 -vd %{buildroot}%{_bindir}
+# Install the binary ourselves, instead of `GOBIN=... make install` because that
+# does extra things we don't need/want.
+install -m 0755 -vp ./mender-artifact %{buildroot}%{_bindir}/
+
+%files
+%{_bindir}/mender-artifact
+%doc README.md
+%license LICENSE
```
lluiscampos, line 57: To be technically correct we need to add all the licenses from the built-in 3rd party code, right? We are not managing that in our deb packages though.

lluiscampos (commented):
Finally we get to honor the name of the repository which from day one was meant to go beyond one type of packaging :raised_hands: 

I would plan this work top-down. This PR does not have a ticket, so I have no context to tell what the further plan is - where are the tests, publishing, docs...

See comments below too :eyes:

lluiscampos:
> It's hard to believe that I broke these by adding a file not used anywhere:

I don't know how much you are following the QA meetings, but we have problems with this pipeline wrt stability. There are few initiatives in place but for now we just deal with having unstable pipeline for the packages.

recipes/mender-artifact/rpm/mender-artifact.spec
vpodzime: Yes, you're right. But as you say, that's something we should fix in general with something like `make license`.

recipes/mender-artifact/rpm/mender-artifact.spec
vpodzime: > Also the author should be [mender@northern.tech](mailto:mender@northern.tech)

I don't agree with this. It should be whoever does the release. And in this case it was me. When a bot does it or it's part of CI/CD, we can use any address of choice.

recipes/mender-artifact/rpm/mender-artifact.spec
vpodzime: > How will be build nightlies for the experimental repo?

That's out of scope of this PR, really. First, I just want to have a spec file that can be used for building RPMs. Even if it was done manually on someone's laptop. Then we can have a repo (or repos). And only then we might consider if we need some repo(s) for nightlies as well.

recipes/mender-artifact/rpm/mender-artifact.spec
vpodzime: An RPM spec can have conditions and lots of logic in it and you can use `rpmbuild --define` or even `rpmbuild --with-something` to parametrize the build. Why should we have multiple spec files? We can use the latter scheme you suggested, but I'd only do that if it turns out to be necessary which I can't see a reason for (famous last words, of course...).

vpodzime:
> I would plan this work top-down. This PR does not have a ticket, so I have no context to tell what the further plan is - where are the tests, publishing, docs...

That's not what this PR is about. It's really about "I want to be able to build an RPM for *mender-artifact*" and really just a side-effect of me requiring such spec to test builds in OBS. I could have put the spec file in a gist or even keep it to myself. But I feel like it's a nice first step towards something more complete and comprehensive, just like what you describe, hence this PR.

lluiscampos:
> I could have put the spec file in a gist or even keep it to myself

I think that is much better, actually. This repo is already quite messy only with the "official" supported packages, adding this just for future reference without proper planning of the epic and support policy is asking for trouble in my opinion (this was one of the conclusions on January workshop, isn't it? That we should not just accept one person wish to get something on a mender repo - I can give examples privately on this)

vpodzime:
> I think that is much better, actually. This repo is already quite messy only with the "official" supported packages, adding this just for future reference without proper planning of the epic and support policy is asking for trouble in my opinion

Works for me. In fact, I can simply close this PR, leaving the spec available here for the future work.

PR #521, opened 2025-04-15
ci: Migrate to priviliged runner for publish jobs

alfrunes:


.gitlab-ci.yml
```diff
@@ -376,7 +376,7 @@ test:acceptance:script:
   stage: publish
   image: ${CI_DEPENDENCY_PROXY_DIRECT_GROUP_IMAGE_PREFIX}/debian:12
   tags:
-    - hetzner-amd-beefy
+    - mender-qa-worker-generic-light
```
oldgiova, line 379: If this is from a protected branch, you can use the hetzner protected runner instead:
```suggestion
    - hetzner-amd-beefy-privileged
```

PR #533, opened 2025-06-25
fix: Require ssh-client for mender-artifact not ssh

vpodzime:
Requiring `ssh` means the SSH client as well as daemon/server is installed, pulling in lots of extra dependencies like Systemd, DBus, etc. `mender-artifact` only requires the `ssh` command, i.e. the client.

Ticket: MEN-8433
Changelog: none


elkoniu (approved):
LGTM

PR #541, opened 2025-07-24
chore: Add an architecture check to armhf packages

vpodzime:
`armhf` can mean both ARMv6 and ARMv7. We build our `armhf` packages for ARMv7 so let's make sure they are not installed on ARMv6 devices.

The easiest and most user-friendly way to perform such a check is in the packages' `preinst` scriptlets that run early in the installation process.

Ticket: QA-1085
Changelog: none


mender-deb-package
```diff
@@ -167,6 +167,35 @@ prepare_recipe() {
     cp support/mender-monitor.service debian/
   fi
 
+  # 'armhf' can mean both ARMv7 (usually and in Debian, in particular) and ARMv6
```
merlin-northern, line 170: wait, in debian armhf always means one thing: armv7.  do we really support raspberry pi that runs armv6?

vpodzime: No, but we used to. So if people try to use our `armhf` packages on 32-bit RPis, we should produce an error.

mender-deb-package
```diff
@@ -167,6 +167,35 @@ prepare_recipe() {
...
+exit 0
+EOF
+    # Some recipes produce multiple packages which have their own scriptlets so
+    # there can be multiple we need to patch.
+    # We need to replace the unindented, and thus unconditional (because we are
+    # good guys), exit 0 with the check followed by a exit 0.
+    for file in debian/*preinst; do
+      sed -ie "/^exit 0/ { r $$.preinst_check
+                         ; d }" "$file"
+    done
```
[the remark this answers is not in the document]
vpodzime: > in general I do not think it is a good idea to patch the preinst here, I would patch the recipies. This will create another plave where some preinst logic is stored.

I agree. But I don't see any way to do a check like this in the recipies. We are cross-compiling/building those packages so the architectures we run the build on doesn't match the architectures of the packages.

So if you see a better way to do this, let me know. I'm not aware of any, as much as I don't like this patching.

PR #543, opened 2025-07-25
ci: Start publishing Debian armhf (ARMv7) packages and binaries

vpodzime:
The Raspberry Pi OS armhf (ARMv6) packages are no longer being published as Debian armhf packages so we can no publish the real Debian armhf (ARMv7) packages.

Ticket: QA-1088
Changelog: none

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -440,17 +440,14 @@ test:acceptance:workstation_tools:packages:
   }
 
 .template:publish:s3:device-components:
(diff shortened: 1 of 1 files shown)


vpodzime:
Packages from the pipeline made into the *experimental* repo and I was able to install them into a Debian 12 ARMv7 container:
```
root@11e67ac4e33c:/# uname -m
armv7l
root@11e67ac4e33c:/# mender-update --version
9fc5b2fe
```

PR #544, opened 2025-07-29
chore: Use the new package repositories

elkoniu:
Old debian packages repository has been migrated
into two separated package repositories:
* workstation-tool
* device-components

From now on packages should be installed from the
new repositories.

Ticket: QA-1096
Changelog: Use the new package repositories

### DIFF
--- a/scripts/install-mender.sh
+++ b/scripts/install-mender.sh
@@ -318,7 +318,12 @@ function get_latest_version_of_commercial_component() {
 }
 
 init() {
-    REPO_URL=https://downloads.mender.io/repos/device-components
+    # This is a deprecated package repository, but we use it for:
+    # * getting GPG key
+    # * detecting deprecated repository link :)
+    REPO_URL=https://downloads.mender.io/repos/debian
+    WORKSTATION_REPO_URL=https://downloads.mender.io/repos/workstation-tools
+    COMPONENTS_REPO_URL=https://downloads.mender.io/repos/device-components
 
     parse_args "$@"
 
@@ -331,9 +336,9 @@ init() {
 
     # Translate Debian "channel" into Mender version for commercial packages
     if [ "$CHANNEL" = "experimental" ]; then
-        VERSION="master"
+        VERSION="stable"
     else
-        VERSION="latest"
+        VERSION="experimental"
     fi
 
     echo "  Installing commercial components from source:"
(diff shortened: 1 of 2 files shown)


scripts/install-mender.sh
```diff
@@ -380,8 +385,14 @@ add_repo() {
         exit 1
     fi
 
-    local repo="deb [arch=$ARCH] $REPO_URL $LSB_DIST/$DIST_VERSION/$CHANNEL main"
-    echo "Installing Mender APT repository at $MENDER_APT_SOURCES_LIST..."
+    # Add workstation-tools repository
+    local repo="deb [arch=$ARCH] $WORKSTATION_REPO_URL $LSB_DIST/$DIST_VERSION/$CHANNEL main"
+    echo "Installing Mender workstation-tools APT repository at $MENDER_APT_SOURCES_LIST..."
+    echo "$repo" > "$MENDER_APT_SOURCES_LIST"
```
michalkopczan, line 391: shouldn't this be >> instead of >?

scripts/install-mender.sh
```diff
@@ -380,8 +385,14 @@ add_repo() {
...
+
+    # Add device-components repository
+    repo="deb [arch=$ARCH] $COMPONENTS_REPO_URL $LSB_DIST/$DIST_VERSION/$CHANNEL main"
+    echo "Installing Mender device-components APT repository at $MENDER_APT_SOURCES_LIST..."
     echo "$repo" > "$MENDER_APT_SOURCES_LIST"
```
michalkopczan, line 396: same here

scripts/tests/test_install_mender_sh.py
```diff
@@ -150,19 +150,33 @@ def local_apt_repo_from_built_packages(container):
...
+def local_apt_repo_from_experimental_packages(container, pool_paths, dest):
     container.run(f"mkdir {dest}")
     for path in pool_paths:
-        url = f"https://downloads.mender.io/repos/debian/pool/test-packages/{path}"
+        url = f""
+        # There are two repositories: workstation-tools and device-components
+        # workstation-tools contains only: mender-artifact, mender-cli
+        # device-components contains everything else
+        if "mender-artifact" in str(path) or "mender-cli" in str(path):
+            url = f"https://downloads.mender.io/repos/device-components/pool/experimental/{path}"
```
michalkopczan, line 177: device-components->workstation-tools

danielskinstad (changes requested):
We're already using the device-components repo here, https://github.com/mendersoftware/mender-dist-packages/commit/91aa25c28e5af0c1a372cf3fa1f3e1eb6e1f2c7e. And we don't need the workstation tools repo here, as this is to be run on a device not a workstation, so we shouldn't need mender-arifact or mender-cli there.

scripts/install-mender.sh
elkoniu: Fixed:)

scripts/tests/test_install_mender_sh.py
elkoniu: Fixed:)

scripts/install-mender.sh
elkoniu: Fixed:)

elkoniu:
Hi @danielskinstad - thanks for the review - I guess I will just close this PR then. Do you think there is anything worth keeping it? The only question I have - in original code we are using VERSION `master` or `latest`. If this is correct in the context of the new repo? I think it should be `stable` and `experimental` instead?

PR #553, opened 2025-09-25
chore: Add Debian 13 support

elkoniu:
Ticket: MEN-8755

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -93,7 +93,7 @@ include:
     matrix:
       - BUILD: crosscompile
         DISTRO: debian
-        RELEASE: [bullseye, bookworm]
+        RELEASE: [bullseye, bookworm, trixie]
         ARCH: [amd64, armhf, arm64]
       - BUILD: crosscompile
         DISTRO: ubuntu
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -421,6 +421,27 @@ test:pkgs:workstation-tools:
     - test -x /usr/bin/mender-cli
     - mender-cli --version
 
+test:pkgs:workstation-tools:trixie
```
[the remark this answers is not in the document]
elkoniu: ;_; <ashemend>

.gitlab-ci.yml
```diff
@@ -421,6 +421,27 @@ test:pkgs:workstation-tools:
     - test -x /usr/bin/mender-cli
     - mender-cli --version
 
+test:pkgs:workstation-tools:trixie:
```
lluiscampos, line 424: This job is exactly like above except for:
* The job name
* `needs`
* `image`

Please consider having a hidden job for example `.test:pkgs:workstation-tools` and then these two jobs extending it. See how it is done elsewhere.

Good initiative adding the tests, though!

lluiscampos (approved):
It looks good. See my comment for how it could look *better*.

.gitlab-ci.yml
elkoniu: I have added the tests with the common part, hope it is according to CI gods rules :D

PR #557, opened 2025-10-01
test: Clean-up traces of workstation tools tests in test framework

lluiscampos:
The tests under `tests/` are in effect device components (only) tests since 7137b13f4.


vpodzime (approved):
Thanks for cleaning up after me! :grin:

elkoniu (approved):

PR #559, opened 2025-10-01
ci: Specify `neeeds` for CI build packages jobs

lluiscampos:
So that they can run despite previous stages failing like static tests.


elkoniu (approved):
LGTM

PR #561, opened 2025-10-01
ci: Use latest Debian version as CI image

elkoniu:
Ticket: None


lluiscampos (approved):
> @lluiscampos I am not sure if this is important but last time we used Debian 12 even for Debian 11 related actions. So my guess is that it shall always be the newest Debian image? Feel free to close this PR as this is not relevant to Debian 13 support.

This is very good, thanks!

These changes only the publish jobs, from which it doesn't matter really which os we use, but using the most recent one will give us more time until we have to update it again.

PR #564, opened 2025-10-03
Deploy candidate

elkoniu:


scripts/install-mender.sh
```diff
@@ -625,6 +628,9 @@ check_dist_and_version() {
                 fi
                 DIST_VERSION="$(sed 's/\/.*//' /etc/debian_version | sed 's/\..*//')"
                 case "$DIST_VERSION" in
+                    13)
+                        DIST_VERSION="trixie"
+                    ;;
                     12)
                         DIST_VERSION="bookworm"
                     ;;
```
lluiscampos, line 636: Looks good!

FYI @elkoniu When doing a `master` -> `production` update the only thing that really gets updated is this script.

lluiscampos (approved):

PR #565, opened 2025-10-03
ci: Update install test script to run on Debian 13

elkoniu:
Ticket: MEN-8756


elkoniu:
@mender-test-bot sync

.gitlab-ci.yml
```diff
@@ -303,19 +303,22 @@ test:check-python3-formatting:
...
     #- mender-qa-worker-generic
     - hetzner-amd-beefy
   needs:
-    # Note that we are only testing packages from Debian bullseye
-    - "build:pkgs:device-components: [crosscompile, debian, bullseye, amd64]"
-    - "build:pkgs:device-components: [crosscompile, debian, bullseye, armhf]"
-    - "build:pkgs:workstation-tools: [crosscompile, debian, bullseye, amd64]"
+    - "build:pkgs:device-components: [crosscompile, debian, amd64]"
+    - "build:pkgs:device-components: [crosscompile, debian, armhf]"
+    - "build:pkgs:workstation-tools: [crosscompile, debian, amd64]"
```
lluiscampos: This is technically wrong, it will never exist such a dependency `build:pkgs:device-components: [crosscompile, debian, amd64]`. It works because all final CI jobs do override `needs`.

Remove `needs` from `.test:pkgs:` hidden CI job (or set to emtpy `[]` as we do in other places, which is also useless but maybe more intuitive for newcomers - you tell me)

lluiscampos (approved):
> From my review, the only outstanding issue is the upgrade tests - requesting changes for this only. I really don't see the benefit of running the upgrade tests from OSs that already install the new package. Can you elaborate on the discussion below so that we can conclude?

Discussed a bit privately. I was wrong in understanding the `test_upgrade_from_*` tests - it is not about what is installed by default in the OS but we do install a given set of static packages.

Sorry for that and thank you for your patience 🙏 

So it makes as it is: we keep testing them only for Debian 11. In the follow-up epic we can see which upgrade paths we can keep testing with Debian 12 and remove the rest.

PR #566, opened 2025-10-03
chore: make mender-orchestrator-support commercial

danielskinstad:
mender-orchestrator-support is a support package for mender-orchestrator, and even though it's open source, it doesn't make sense to install it separately. Make it commercial so we can download the packages from same link as mender-orchestrator

Ticket: MEN-7252


elkoniu (commented):
Why not have it in both places? ;)

elkoniu:
> @danielskinstad Looks good!
> 
> > Why not have it in both places? ;)
> 
> @elkoniu We discussed this in person, sorry that you were left out. The summary is:
> * On the one hand, it does not give any value having the packages published to the public APT repo because it is not installable (they require mender-orchestrator-core which is private)
> * On the other hand simplifies user's instructions to get the packages if all of them (closed an open source) are in the same S3 bucket / same URL prefix

ACK

PR #568, opened 2025-10-06
ci: use k8s runners by default

oldgiova:
To avoid using public shared runners

Ticket: SEC-1133


elkoniu:
@oldgiova does this PR still apply on the latest code or we shall drop it?

oldgiova:
Right, this is not relevant anymore (oct 2025!)

PR #627, opened 2026-06-15
Remove Ubuntu 22 support in favor of new Ubuntu 26 support

elkoniu:
Add support for Ubuntu 26 and remove support for Ubuntu 22

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -194,7 +194,7 @@ include:
         RELEASE: [bookworm, trixie]
         ARCH: [amd64, armhf, arm64]
       - DISTRO: ubuntu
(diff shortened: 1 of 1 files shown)


elkoniu:
Blocked by: https://github.com/mendersoftware/mender-test-containers/pull/718

vpodzime (approved):
I'd remove the _Ubuntu26_ scope from the first commit message. But looks good to me otherwise.

elkoniu:
Thanks for the review @vpodzime . I have used the same scope in other PRs to easier identify them in the future. So if someone will be adding Ubuntu LTS in the future - will be able to filter out those commits from the git history even without access to original JIRA items.

PR #628, opened 2026-06-15
chore: use new format for builder image

danielskinstad:
mender-test-containers changed how it published container images. Use the new format.


elkoniu (approved):
LGTM - thanks for fixing this \o/

PR #629, opened 2026-06-15
chore(Ubuntu 26): Add Ubuntu Resolute pkg build missing dependency

elkoniu:
Ticket: MEN-9727

### DIFF
--- a/recipes/mender-orchestrator/debian-master/control
+++ b/recipes/mender-orchestrator/debian-master/control
@@ -4,7 +4,7 @@ Section: admin
 Priority: optional
 Standards-Version: 3.9.2
 Build-Depends-Indep: cmake, debhelper (>= 10)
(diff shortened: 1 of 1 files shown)


elkoniu:
@vpodzime this is missing dependency for device-components build but it is specific to new Ubuntu 25 Resolute. I am wondering if we shall handle it here, or make it part of the Resolute container image.

Making it a container image part sounds like a cleaner solution, on the other hand:
https://github.com/mendersoftware/mender-test-containers/pull/719

On the other hand having it here makes it easier to navigate / modify as we already installing package specific dependencies.

mender-deb-package
```diff
@@ -247,6 +247,11 @@ build_orig() {
 install_build_dependencies() {
   # Install the dependencies from the debian/control file
   apt-get update
+
+  if [ "${OS_CODENAME}" == "resolute" ]; then
+      apt-get install -yq libicu-dev
+  fi
```
vpodzime: This should be a build dependency in the package recipe. But it looks strange, we don't use this library. What exactly failed and how?

vpodzime (approved):
ACK

elkoniu:
The followup investigation on why we need this library in the 1st place will continue in MEN-9868

PR #635, opened 2026-06-19
chore: Change to new commit linter

rewanrashid-boop:
Replace old commit linter with the new one in mendertesting/templates

Ticket: QA-1561


mzedel (changes requested):
you need to also remove the `test:check-commits:` configuration further down in the file 👌

mzedel (approved):

